PR #4, opened 2022-04-07
feat(core): added new tooltip component

vardidzel:


core/src/components/cat-tooltip/cat-tooltip.spec.tsx
```diff
@@ -0,0 +1,18 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { CatTooltip } from './cat-tooltip';
+
+describe.skip('cat-tooltip', () => {
```
fynnfeldpausch, line 4: If possible, dont skip but make it work

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,68 @@
...
+  shadow: true,
+})
+export class CatTooltip {
+  private tooltip?: HTMLElement;
+  private trigger?: Element;
+
+  /**
+   * The content of the tooltip
+   */
+  @Prop() content!: string;
```
fynnfeldpausch, line 16: Don't use `!` for properties. We cannot make sure that the property is actually given by the developer. Instead, you need to handle it in code

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,68 @@
...
+   * Specifies that the tooltip should be disabled. A disabled tooltip is unusable,
+   * and invisible. Corresponds with the native HTML disabled attribute.
+   */
+  @Prop() disabled = false;
+
+  /**
+   * The placement of the tooltip.
+   */
+  @Prop() placement: Placement = 'top';
+
```
fynnfeldpausch, line 28: I forgot two more properties to adjust the delay: `showDelay` and `hideDelay`. Please add them

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,68 @@
...
+  componentDidLoad(): void {
+    if (this.trigger && this.tooltip) {
+      autoUpdate(this.trigger, this.tooltip, () => this.update());
+    }
+  }
+
+  render() {
+    return (
+      <Host>
+        <div
```
fynnfeldpausch, line 38: Is it possible to get rid of the div wrapper and directly use the Host element?

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,68 @@
...
+          aria-valuetext={this.content}
+          role="tooltip"
```
fynnfeldpausch, line 40: Looking at a couple of resources, the aria tooltip role is actually discouraged for tooltips :D. See: https://sarahmhigley.com/writing/tooltips-in-wcag-21/#:~:text=role%3D%22tooltip%22%3A%20the,with%20aria%2Ddescribedby. I think we should follow the approach of angular Material closely. Also with adding a hidden content: https://material.angular.io/components/tooltip/overview#accessibility

core/src/components/cat-tooltip/cat-tooltip.tsx
fynnfeldpausch: Please make sure that the tooltip is properly updated if the content changes while the tooltip is shown

fynnfeldpausch (changes requested):

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+@Component({
+  tag: 'cat-tooltip',
+  styleUrl: 'cat-tooltip.scss',
+  shadow: true,
+})
+export class CatTooltip {
+  private tooltip?: HTMLElement;
+  private trigger?: Element;
+  private showTimeout: NodeJS.Timeout;
```
fynnfeldpausch, line 12: The type is wrong. it should be number. You will get the correct type, if you use the proper window.setTimeout and window.clearTimeout. These are coming from lib dom

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+import {autoUpdate, computePosition, flip, offset, Placement} from "@floating-ui/dom";
+
+@Component({
+  tag: 'cat-tooltip',
+  styleUrl: 'cat-tooltip.scss',
+  shadow: true,
+})
+export class CatTooltip {
+  private tooltip?: HTMLElement;
+  private trigger?: Element;
```
fynnfeldpausch, line 11: should be an HTMLElement also

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+    return (
+      <Host>
+        <div class="tooltip-content" ref={el => (this.trigger = el)}>
+          <slot/>
+        </div>
+        {
+          this.content && !this.disabled &&
+          <div
+            ref={el => (this.tooltip = el)}
+            aria-describedby={this.content}
```
fynnfeldpausch, line 65: ARIA-describedby expects an ID and not a text

fynnfeldpausch: See menu component on how to autogenerate unique IDs

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  disconnectedCallback(): void {
+    this.trigger?.removeEventListener('mouseenter', this.showListener.bind(this));
+    this.trigger?.removeEventListener('mouseleave', this.hideListener.bind(this));
+  }
+
+  render() {
+    return (
+      <Host>
```
fynnfeldpausch, line 57: technically, you could get rid of the host element, as you are not using it

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+        }
+      </Host>
+    );
+  }
+
+  private update() {
+    if (this.trigger && this.tooltip) {
+      computePosition(this.trigger, this.tooltip, {
+        placement: this.placement,
+        middleware: [offset(4), flip()]
```
fynnfeldpausch, line 79: place offset in a static readonly property (see menu component)

core/src/components/cat-tooltip/cat-tooltip.scss
```diff
@@ -0,0 +1,24 @@
+:host {
+  display: contents;
+}
+
```
fynnfeldpausch, line 4: add a disabled selector as well for best practices (see other components)

core/src/components/cat-tooltip/cat-tooltip.scss
```diff
@@ -0,0 +1,24 @@
+:host {
```
fynnfeldpausch, line 1: Time to use the design tokens in this file. start by importing
@use 'variables' as *;
@use 'mixins' as *;

core/src/components/cat-tooltip/cat-tooltip.scss
```diff
@@ -0,0 +1,24 @@
...
+  display: contents;
+}
+
+.tooltip {
+  position: absolute;
+  padding: 5px 10px;
+  background-color: rgba(0, 0, 0, .7);
+  border-radius: 4px;
```
fynnfeldpausch, line 9: border-radius: cat-border-radius('s');

core/src/components/cat-tooltip/cat-tooltip.scss
```diff
@@ -0,0 +1,24 @@
+:host {
+  display: contents;
+}
+
+.tooltip {
+  position: absolute;
+  padding: 5px 10px;
+  background-color: rgba(0, 0, 0, .7);
```
fynnfeldpausch, line 8: needs to be added to the tokens project. I'd go with _color.ui.background.tooltip_ and _color.ui.font.tooltip_ and _opacity.tooltip_. Then use these consistently in the style sheet via `cat-token('color.ui....');`

core/src/components/cat-tooltip/cat-tooltip.scss
```diff
@@ -0,0 +1,24 @@
...
+  border-radius: 4px;
+  color: #fff;
```
fynnfeldpausch, line 10: use token here (see above)

core/src/components/cat-tooltip/cat-tooltip.scss
```diff
@@ -0,0 +1,24 @@
+:host {
+  display: contents;
+}
+
+.tooltip {
+  position: absolute;
```
fynnfeldpausch, line 6: add `@include cat-body('s');`

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  disconnectedCallback(): void {
+    this.trigger?.removeEventListener('mouseenter', this.showListener.bind(this));
+    this.trigger?.removeEventListener('mouseleave', this.hideListener.bind(this));
+  }
+
+  render() {
+    return (
+      <Host>
+        <div class="tooltip-content" ref={el => (this.trigger = el)}>
```
fynnfeldpausch, line 58: Shouldnt this rather be called tooltip-trigger? And it should also have display: contents, no?

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * The delay time for showing tooltip in ms
+   */
+  @Prop() showDelay = 500;
+
+  /**
+   * The delay time for hiding tooltip in ms
+   */
+  @Prop() hideDelay = 500;
```
fynnfeldpausch, line 39: default should be 0

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * The placement of the tooltip.
+   */
+  @Prop() placement: Placement = 'top';
+
+  /**
+   * The delay time for showing tooltip in ms
+   */
+  @Prop() showDelay = 500;
```
fynnfeldpausch, line 34: Default should be 1000

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
...
+   * The delay time for hiding tooltip in ms
+   */
+  @Prop() hideDelay = 500;
+
+  componentDidLoad(): void {
+    if (this.trigger && this.tooltip) {
+      autoUpdate(this.trigger, this.tooltip, () => this.update());
+    }
+    this.trigger?.addEventListener('mouseenter', this.showListener.bind(this));
+
```
fynnfeldpausch, line 46: remove newline

core/src/components/cat-tooltip/cat-tooltip.tsx
```diff
@@ -0,0 +1,104 @@
+import {Component, h, Host, Prop} from '@stencil/core';
+import {autoUpdate, computePosition, flip, offset, Placement} from "@floating-ui/dom";
+
+@Component({
```
fynnfeldpausch, line 4: Please take a look at both a11y concerns mentioned in the angular tooltip library and properly implement them here:

> MatTooltip adds an aria-describedby description that provides a reference to a visually hidden element containing the tooltip's message. This provides screenreaders the information needed to read out the tooltip's contents when the end-user focuses on tooltip's trigger. The element referenced by aria-describedby is not the tooltip itself, but instead an invisible copy of the tooltip content that is always present in the DOM.
> 
> Avoid interactions that exclusively show a tooltip with pointer events like click and mouseenter. Always ensure that keyboard users can perform the same set of actions available to mouse and touch users.

fynnfeldpausch (changes requested):

fynnfeldpausch:
Long press is not working as expected, but I will merge this now and create a follow up ticket to make progress on this.

PR #5, opened 2022-04-07
feat(core): added scrollable component

manyx:


vardidzel (changes requested):
Please add `idea` files in `.gitignore`

core/src/components/cat-scrollable/cat-scrollable.tsx
```diff
@@ -0,0 +1,31 @@
...
+})
+export class CatScrollable {
+
+  /** Flags to enable/disable scroll shadows. */
+  @Prop()
+  shadow: { x?: boolean; y?: boolean; } = { x: true, y: true };
+
+  /** Flags to enable/disable overflow. */
+  @Prop()
+  overflow: { x?: boolean; y?: boolean; } = { y: true };
```
vardidzel, line 16: Please move this  `{ x?: boolean; y?: boolean; }`  in another file.

core/src/components/cat-scrollable/cat-scrollable.tsx
```diff
@@ -0,0 +1,31 @@
...
+@Component({
+  tag: 'cat-scrollable',
+  styleUrl: 'cat-scrollable.css',
+  shadow: true,
+})
+export class CatScrollable {
+
+  /** Flags to enable/disable scroll shadows. */
+  @Prop()
+  shadow: { x?: boolean; y?: boolean; } = { x: true, y: true };
```
vardidzel, line 12: same here

fynnfeldpausch: as far as I read in best practices, it is discouraged to use complex objects in properties. So I suggest to go with individual props like shadowX and shadowY

core/src/components/cat-scrollable/cat-scrollable.tsx
fynnfeldpausch: same here

PR #6, opened 2022-04-11
Feat/scrollable

manyx:


core/src/components/cat-scrollable/cat-scrollable.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <div
+          ref={el => (this.scrollElement = el)}
+          class='scrollable-content'>
+          <slot></slot>
+        </div>
+      </Host>
+    );
+  }
+
+  private attachEmitter(from: 'top' | 'left' | 'right' | 'bottom', emitter: EventEmitter<void>, buffer: number) {
```
vardidzel, line 115: no return types for private methods.

vardidzel: Also, please avoid the use of the internal type/interfaces, move `'top' | 'left' | 'right' | 'bottom'` separately.

vardidzel (changes requested):

core/src/components/cat-scrollable/cat-scrollable.tsx
```diff
@@ -0,0 +1,151 @@
...
+  tag: 'cat-scrollable',
+  styleUrl: 'cat-scrollable.scss',
+  shadow: true
+})
+export class CatScrollable {
+  scrollElement?: HTMLElement;
+  scrollWrapperElement?: HTMLElement;
+  private readonly init = new Subject<void>();
+  private readonly destroyed = new Subject<void>();
+  private scrolled!: Observable<Event>;
```
fynnfeldpausch, line 15: get rid of the `!` here

fynnfeldpausch (changes requested):

core/src/components/cat-scrollable/cat-scrollable.tsx
manyx: it doesn't work without !

fynnfeldpausch (approved):

PR #7, opened 2022-04-13
feat(core): test without prepare command

vardidzel:


.github/workflows/core.yml
```diff
@@ -18,11 +18,15 @@ jobs:
       - name: Setup
         uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: '16.13.1'
```
fynnfeldpausch, line 21: Is there any reason to pin the version other than "best practice"? Or was there an error caused by this? I would like to keep the more generic version if possible. We have also not pinned it in the nvm file and in the package.json

.github/workflows/core.yml
```diff
@@ -18,11 +18,15 @@ jobs:
...
           cache: 'npm'
           cache-dependency-path: core/package-lock.json
       - name: Install and Build
         run: |
+          npm run prebuild:tokens && npm run build:tokens
```
fynnfeldpausch, line 26: I don't quite understand why it is not working without the linking of the projects. And while this is a solution, I think, it is wrong to do it. Because, we can now build a version of core, that relies on the linked tokens or icons, even if those are not yet published. And the build will succeed. IMO, the build should use fixed versions of tokens & icons and not rely on npm link

fynnfeldpausch (changes requested):

fynnfeldpausch (approved):

PR #9, opened 2022-04-19
feat(core): changes for coyo app integration

manyx:


angular/projects/catalyst/src/lib/catalyst.module.ts
```diff
@@ -17,7 +21,7 @@ const CatComponents = [
 export class CatalystModule {
 
   static forRoot(): ModuleWithProviders<CatalystModule> {
-    defineCustomElements();
+    defineCustomElements(window);
```
vardidzel, line 24: I'm just wondering if can `defineCustomElements` be called without `window`?

angular/projects/catalyst/package.json
```diff
@@ -11,6 +11,7 @@
   },
   "peerDependencies": {
     "@angular/core": ">=12.0.0",
-    "rxjs": ">=6.6.0"
+    "rxjs": ">=6.6.0",
+    "@haiilo/catalyst": ">=0.0.1"
```
fynnfeldpausch, line 15: Thinking about it, I guess it is not a peerDependency, as the core and angular wrapper a closely coupled. So it would rather be a real dependency with a fixed version. Upside: No need to specify the dependency on COYO side anymore.

angular/projects/catalyst/src/lib/catalyst.module.ts
```diff
@@ -5,7 +5,11 @@ import * as Components from './directives/proxies';
 const CatComponents = [
   Components.CatButton,
   Components.CatIcon,
-  Components.CatSpinner
+  Components.CatSpinner,
```
fynnfeldpausch, line 8: Was this autogenerated?

angular/projects/catalyst/src/lib/catalyst.module.ts
fynnfeldpausch: Same question. The examples I saw were calling the function without window?

angular/tsconfig.json
```diff
@@ -5,8 +5,8 @@
     "baseUrl": "./",
     "paths": {
       "catalyst": [
-        "dist/catalyst/catalyst",
-        "dist/catalyst"
+        "dist/catalyst",
+        "dist/catalyst/*"
```
fynnfeldpausch, line 9: Was this autogenerated?

core/src/components/cat-button/cat-button.scss
```diff
@@ -11,8 +11,10 @@
 .cat-button {
   font: inherit;
   display: flex;
+  flex: 1 1 auto;
```
fynnfeldpausch, line 14: please revert this change. we will fix it separately

core/src/components/cat-skeleton/readme.md
```diff
@@ -15,6 +15,15 @@
 | `variant` | `variant` | The rendering style of the skeleton.                                                                                                                   | `"body" \| "circle" \| "head" \| "rectangle" \| "square"` | `'rectangle'` |
 
 
+## CSS Custom Properties
```
fynnfeldpausch, line 18: auto-generated?

fynnfeldpausch (changes requested):

angular/projects/catalyst/src/lib/catalyst.module.ts
manyx: I guess it can work without it, I saw some examples that used window so I added it but I'll remove it.

angular/projects/catalyst/src/lib/catalyst.module.ts
manyx: No, I added them to include the new components that we added.

core/src/components/cat-button/cat-button.scss
manyx: done

core/src/components/cat-skeleton/readme.md
manyx: yes

angular/projects/catalyst/package.json
manyx: moved it to dependencies.

angular/projects/catalyst/src/lib/catalyst.module.ts
fynnfeldpausch: ah, makes sense. i thought the stencil CLI already takes care of it but apparently it does not :D

PR #14, opened 2022-04-27
PMI-17: notifications

manyx:


core/src/styles/index.scss
```diff
@@ -25,3 +25,8 @@
 @use 'utils/typography';
 @use 'utils/visibility';
 @use 'utils/ratio';
+@use 'utils/toast';
+
+@import  '../../node_modules/toastify-js/src/toastify.css';
```
fynnfeldpausch, line 30: use @use instead of import and place it in a new // -- Vendor section between Reset and Core

core/src/styles/utils/_toast.scss
```diff
@@ -0,0 +1,72 @@
+@import 'variables';
```
fynnfeldpausch, line 1: use @use

core/src/styles/utils/_toast.scss
```diff
@@ -0,0 +1,72 @@
...
+@import 'mixins';
+
+.toastify {
```
fynnfeldpausch, line 4: do you need this to increase selector specificity? otherwise just go with .cat-toastify

core/src/utils/notifications-service.ts
```diff
@@ -0,0 +1,122 @@
+import Toastify, { Offset } from 'toastify-js'
+
+export enum TypeIcons {
+  'success' = 'check-circle-filled',
+  'error' = 'sparkle-filled',
+  'info' = 'sparkle-filled'
+}
+
+export interface ToastOptions {
+  text?: string | undefined;
```
fynnfeldpausch, line 10: remove undefined and ? from all types here and in the implementation use the type Partial<ToastOptions>

core/src/utils/notifications-service.ts
```diff
@@ -0,0 +1,122 @@
+import Toastify, { Offset } from 'toastify-js'
+
+export enum TypeIcons {
+  'success' = 'check-circle-filled',
+  'error' = 'sparkle-filled',
+  'info' = 'sparkle-filled'
+}
+
+export interface ToastOptions {
```
fynnfeldpausch, line 9: Add docs to all options

fynnfeldpausch, line 9: We want to expose as little options as possible to a) keep it very simple, b) make it easier to switch to another library and c) make it opinionated and disallow too many variations. So please remove the following options and find sensible defaults:

- text / node should become `content: string | Node` IMO (yes, we then need some kind of mapping logic for the options).
- duration
- selector
- destination
- newWindow
- gravity / position should be combined into one position attribute, that takes stuff like top-left, top-right... 
- className
- stopOnFocus (should be true)
- callback
- offset
- style
- oldestFirst

fynnfeldpausch: yes, there is not much left, but that is intentional :D

core/src/utils/notifications-service.ts
```diff
@@ -0,0 +1,122 @@
...
+        ${hasMessage ? `<div class="cat-toastify-message">${message}</div>` : ''}
+      </div>`;
+    return template.content.firstChild as HTMLElement;
+  }
+
+  error(title: string, message = ' ', options?: ToastOptions): void {
+    options = {
+      ...{
+        node: this.toastHTMLTemplate(title, message, options),
+        duration: 300000,
```
fynnfeldpausch, line 80: default duration should be a private static constant on the top of the class. it should be 8s.

fynnfeldpausch (changes requested):

fynnfeldpausch (approved):

PR #15, opened 2022-04-28
Feat/radio

geralrodrigues:


core/src/components/cat-radio/cat-radio.tsx
```diff
@@ -0,0 +1,89 @@
...
+  styleUrl: 'cat-radio.scss',
+  shadow: true
+})
+export class CatRadio {
+  private readonly id = `cat-radio-${++nextUniqueId}`;
+
+  /**
+   * Whether this radio is checked.
+   * */
+  @Prop() checked?: boolean;
```
fynnfeldpausch, line 20: for boolean props, please assign a default value instead of making it optional

core/src/components/cat-radio/cat-radio.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * The value of the radio component.
+   */
+  @Prop() value?: string;
+
+  /**
+   * The label of the radio that is visible.
+   */
+  @Prop() label?: string;
```
fynnfeldpausch, line 45: label should default to an empty string

core/src/components/cat-radio/cat-radio.tsx
```diff
@@ -0,0 +1,89 @@
...
+          class="cat-radio-input"
+          checked={this.checked}
+          required={this.required}
+          disabled={this.disabled}
+          name={this.name}
+          value={this.value}
+          onChange={this.onChange.bind(this)}
+        />
+        <span class="cat-radio-checked-circle" />
+        <label htmlFor={this.id} part="label" class="cat-radio-label" aria-label={this.label}>
```
fynnfeldpausch, line 79: - remove aria label from the label
- always include the label content in the label element
- properly hide the label if hideLabel is set (possibly using the visibility-hidden mixin, we provide)

fynnfeldpausch (changes requested):

core/src/components/cat-radio/cat-radio.tsx
fynnfeldpausch, line 45: We want to put accessibility first, so one of the approaches I took was that components actively complain if they are not accessible. So what do you think of logging an error if no label is set (because it always needs to be set, even if it is hidden). Take a look at the `componentWillRender` method in the `cat-button` component for example and implement a similar behavior => log error if label is empty and component is rendered)

fynnfeldpausch (approved):

PR #18, opened 2022-04-29
Fix/tooltip

geralrodrigues:
Added an event in the window so that when you press anywhere on the device the tooltip will disappear.

Modified the touchend event of the trigger so that the short taps work, that is, it removes the touchTimeout only when it exists.

And modified the touchstart of the trigger so that the window event is not propagated, so you can open the tooltip in the long press.

Solve the ACs of ticket PMI-57


fynnfeldpausch:
Hey @geralrodrigues. I am a bit confused because I was expecting other changes. Can you add some description to the PR what the changes are and why they were necessary, pls.

fynnfeldpausch (changes requested):
see comments

fynnfeldpausch:
Also, there is another problem with tooltips: change the tooltip target to a button and hover over it on your desktop computer. the button receives focus purely by hovering over it. that is not wanted.

https://user-images.githubusercontent.com/1246913/170223298-176ffe8c-f83a-4ee0-90a3-2c28fa398e82.mov

fynnfeldpausch (approved):

PR #22, opened 2022-05-10
Feat/modal

geralrodrigues:


core/src/components/cat-modal/cat-modal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      });
+    }
+  }
+
+  /**
+   * Shows the modal.
+   */
+  @Method()
+  async show() {
+    this.modalWrapper?.classList.add('is-visible');
```
vardidzel, line 44: I think a better way is to have a state of the `show` and use that state in a `class` attribute
for example: `class={{show: this.show}}`
This is a much cleaner way IMHO

geralrodrigues: Ok, I'll take a look at it, it also seems to me that it would be cleaner, thanks for commenting 😊

fynnfeldpausch:
Hey @geralrodrigues, nice work. I still do see some a11y problems with the implementation. Even though you have put in a lot of effort, I'd like to discuss another option, that I found. What do you think of https://a11y-dialog.netlify.app/ ? It is a very lightweight implementation of a modal with the focus of a11y. We could use it as a dependency and build a wrapper around it. Do you think this would yield a better result and also reduce maintenance costs in the long run?

geralrodrigues:
> Hey @geralrodrigues, nice work. I still do see some a11y problems with the implementation. Even though you have put in a lot of effort, I'd like to discuss another option, that I found. What do you think of https://a11y-dialog.netlify.app/ ? It is a very lightweight implementation of a modal with the focus of a11y. We could use it as a dependency and build a wrapper around it. Do you think this would yield a better result and also reduce maintenance costs in the long run?

PR #25, opened 2022-05-16
Fix/boolean props

geralrodrigues:


vardidzel (approved):

core/src/components/cat-button/cat-button.tsx
```diff
@@ -78,9 +78,9 @@ export class CatButton {
   @Prop() submit = false;
 
   /**
-   * Ellipse overflowing button content.
+   * Disables ellipse overflowing button content.
    */
-  @Prop() ellipsed = true;
+  @Prop() noEllipsed = false;
```
fynnfeldpausch, line 83: pls rename to `noEllipsis`

fynnfeldpausch (changes requested):
please rename the one prop

fynnfeldpausch (approved):

PR #26, opened 2022-05-17
feat(core) added select

vardidzel:


fynnfeldpausch:
- [ ] Simplify placeholder Prop. there should be only one and that prop should set the placeholder.
- [ ] remove render choice limit for now
- [ ] fuse options need to be gone entirely. basically, we want this simple behavior: search input is split by whitespace, then perform a case-insensitive contains search and combine results using AND
- [ ] component needs to support a disabled input prob. methods for enable and disable should be removed
- [ ] searchresultlimit should be set to some high value (e.g. 24) and Prop is removed
- [ ] options that have the default value should be removed from the settings object

core/src/components/cat-select/variables.scss
```diff
@@ -0,0 +1,4 @@
+@use 'variables' as *;
```
fynnfeldpausch, line 1: remove this file.

core/src/components/cat-select/variables.scss
```diff
@@ -0,0 +1,4 @@
...
+@use 'mixins' as *;
+
+$choices-bg-color: #fff;
```
fynnfeldpausch, line 4: use `cat-token('color.ui.background.body');` instead

core/src/components/cat-select/cat-select.tsx
```diff
@@ -0,0 +1,314 @@
...
+  @Prop() position: 'auto' | 'top' | 'bottom' = 'auto';
+  @Prop() searchable = true;
+  @Prop() disabled = false;
+
+  // tags?
+  @Prop() addItems = true;
+  @Prop() editItems = false;
+  @Prop() paste = false;
+
+  @Prop() choices?: Option[];
```
fynnfeldpausch, line 55: We still need to get the number of Props down to simplify the component. I will arrange a discussion meeting.

core/src/components/cat-select/interfaces.ts
```diff
@@ -0,0 +1,62 @@
+export type ClassNames = {
```
fynnfeldpausch, line 1: any possibility to not define this ourselves?

fynnfeldpausch:
Bildschirmfoto 2022-06-21 um 11 28 18

undefined shows as description in all selects

fynnfeldpausch (changes requested):

fynnfeldpausch (approved):
we will push issues to follow up tickets

PR #27, opened 2022-05-26
Feat/tabs

geralrodrigues:


core/src/components/cat-tab/cat-tab.scss
```diff
@@ -0,0 +1,6 @@
+@use 'variables' as *;
+@use 'mixins' as *;
+
+:host {
+  display: block;
```
fynnfeldpausch, line 5: as seen in other components, add a hidden case:

```
  &[hidden] {
    display: none;
  }
```

core/src/components/cat-tab/cat-tab.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  @Listen('click')
+  onClick(event: MouseEvent) {
+    this.tabClick.emit(event);
+  }
+
+  render() {
+    return (
+      <Host>
+        <slot></slot>
```
fynnfeldpausch, line 53: I don't get your idea behind this slot here. Why place the button implementation in the tabs component and not directly in here?

core/src/components/cat-tab/cat-tab.tsx
```diff
@@ -0,0 +1,57 @@
+import { Component, h, Prop, Event, EventEmitter, Listen, Host } from '@stencil/core';
+import { Breakpoint } from '../../utils/breakpoints';
+
+@Component({
+  tag: 'cat-tab',
+  styleUrl: 'cat-tab.scss',
+  shadow: true
+})
+export class CatTab {
```
fynnfeldpausch, line 9: a tab needs to be able to be deactivated...

core/src/components/cat-tabs/cat-tabs.scss
```diff
@@ -0,0 +1,32 @@
...
+  box-shadow: inset 0 -1px 0 0 cat-token('color.ui.border.tab');
+  padding: 0.5rem;
+}
+
+.tab-active {
+  box-shadow: inset 0 -3px 0 0 cat-token('color.ui.border.activeTab');
+}
+
+.tab-align-left::part(button) {
+  justify-content: start;
```
fynnfeldpausch, line 23: there is a misunderstanding what the tab alignment should look like. Here you can see an example of left and center aligned tabs: https://stackblitz.com/run?file=src%2Fapp%2Ftab-group-align-example.ts
Here is an example of justified tabs: https://stackblitz.com/run?file=src%2Fapp%2Ftab-group-stretched-example.ts

so the alignment is applied to the tabs and not to the content of the tabs. makes sense? Please get in touch if this is still unclear. If you want, you can also add a right alignment option (as seen in the first example). But this would be more for the sake of completeness

core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  @Watch('activeTabId')
+  onActiveTabChanged(newActiveTab: string): void {
+    const activeTab = this.tabs.find(value => value.id === newActiveTab);
+    activeTab?.click();
+  }
+
+  componentWillLoad(): void {
+    this.tabs = Array.from(this.hostElement.querySelectorAll('cat-tab'));
+    if (this.tabs.length > 0) {
```
fynnfeldpausch, line 37: if (this.tabs.length)

core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -0,0 +1,107 @@
...
+      let nextUniqueId = 0;
+      this.tabs.forEach(tab => {
+        if (!tab.id) tab.id = `cat-checkbox-${nextUniqueId++}`;
```
fynnfeldpausch, line 40: copy paste error. should not be checkbox id

core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -0,0 +1,107 @@
...
+  @Watch('activeTabId')
+  onActiveTabChanged(newActiveTab: string): void {
+    const activeTab = this.tabs.find(value => value.id === newActiveTab);
+    activeTab?.click();
+  }
+
+  componentWillLoad(): void {
+    this.tabs = Array.from(this.hostElement.querySelectorAll('cat-tab'));
+    if (this.tabs.length > 0) {
+      let nextUniqueId = 0;
```
fynnfeldpausch, line 38: defining this here will lead to non-unique IDs for different tabs components. we need to push it o the outside of the component just like the other examples

core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -0,0 +1,107 @@
...
+      this.tabs.forEach(tab => {
+        if (!tab.id) tab.id = `cat-checkbox-${nextUniqueId++}`;
+      });
+      this.activeTabId = this.activeTab;
+    }
+  }
+
+  componentDidRender(): void {
+    this.buttons.forEach((button: HTMLCatButtonElement) => {
```
fynnfeldpausch, line 47: explicit typing not required

core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -0,0 +1,107 @@
...
+      const element = button.shadowRoot?.querySelector('button');
+      element?.setAttribute('tabindex', button.buttonId === this.activeTabId ? '0' : '-1');
+    });
+  }
+
+  @Listen('keydown')
+  onKeydown(event: KeyboardEvent): void {
+    if (['ArrowDown', 'ArrowUp', 'ArrowRight', 'ArrowLeft'].includes(event.key)) {
```
fynnfeldpausch, line 55: nice implementation. Works. Did you check if one could use the focus trap library (that we already using), can be used here?

fynnfeldpausch (changes requested):

core/src/components/cat-tab/cat-tab.tsx
fynnfeldpausch: aka disabled

core/src/components/cat-tabs/cat-tabs.tsx
geralrodrigues: in this case it would not be possible to use the focus trap. It only works with the Tab and Shift+Tab keys, not the arrow keys.

fynnfeldpausch (approved):

PR #36, opened 2022-06-29
feat(core): fix value accessor

manyx:
fixed value accessor bugs


angular/projects/catalyst/src/lib/catalyst.module.ts
```diff
@@ -2,8 +2,16 @@ import { InjectionToken, ModuleWithProviders, NgModule } from '@angular/core';
 import { CatI18nRegistry, CatIconRegistry } from '@haiilo/catalyst';
 import { defineCustomElements } from '@haiilo/catalyst/loader';
 import * as Components from './directives/proxies';
+import { TextValueAccessor } from "./directives/text-value-accessor";
+import { SelectValueAccessor } from "./directives/select-value-accessor";
+import { RadioValueAccessor } from "./directives/radio-value-accessor";
+import { BooleanValueAccessor } from "./directives/boolean-value-accessor";
 
 const CatComponents = [
+  TextValueAccessor,
```
vardidzel, line 11: Please, create one variable with the name CatDirectives and move all directives inside.
After, just add it in declarations this way: `declarations: [...CatComponents, ...CatDirectives]`

PR #38, opened 2022-06-30
Feat/radio

geralrodrigues:


core/src/components/cat-radio-group/cat-radio-group.tsx
```diff
@@ -0,0 +1,76 @@
...
+  @Prop() name?: string;
+
+  /**
+   * Whether this radio group is disabled.
+   */
+  @Prop() disabled = false;
+
+  componentDidLoad(): void {
+    this.catRadioGroup = Array.from(this.hostElement.querySelectorAll(`cat-radio`));
+    this.catRadioGroup.forEach(catRadio => {
```
fynnfeldpausch, line 25: this will not work if the property is changed later. Or toggled dynamically. We need to execute it on watch of the property as well as on load. Probably applies to both the disabled and the name.

core/src/components/cat-radio-group/cat-radio-group.tsx
```diff
@@ -0,0 +1,76 @@
...
+      const targetIdx = activeIdx < 0 ? 0 : (activeIdx + activeOff + targetElements.length) % targetElements.length;
+      targetElements[targetIdx].setFocus();
+      targetElements[targetIdx].shadowRoot?.querySelector('input')?.click();
+      this.updateTabIndex();
+      event.preventDefault();
+    }
+  }
+
+  @Listen('click')
+  onClick(event: MouseEvent): void {
```
fynnfeldpausch, line 50: do we need to check if it is disabled and then do an early return?

core/src/components/cat-radio-group/cat-radio-group.tsx
```diff
@@ -0,0 +1,76 @@
...
+    if (catRadioElement && catRadioElement.checked) {
+      const catRadioElements = this.catRadioGroup.filter(value => value !== catRadioElement);
+      catRadioElements.forEach(value => (value.checked = false));
+      this.updateTabIndex();
+    }
+  }
+
+  render() {
+    return (
+      <Host role="radiogroup">
```
fynnfeldpausch, line 61: as far as I read, setting attributes and classes on the host element is bad practices. What do you think of using something like this?

    
    return (
      <div role="radiogroup">
        <slot></slot>
      </div>
    );

But we would then also need an additional property `@Prop({ attribute: 'a11y-label' }) a11yLabel?: string;` to set the aria label on the div... This would also be more consistent to the button element for example

fynnfeldpausch (changes requested):

core/src/components/cat-radio-group/cat-radio-group.tsx
geralrodrigues: By changing it to onInput, it would no longer be necessary

core/src/components/cat-radio-group/cat-radio-group.tsx
```diff
@@ -0,0 +1,96 @@
...
+    if (catRadioElement && catRadioElement.checked) {
+      const catRadioElements = this.catRadioGroup.filter(value => value !== catRadioElement);
+      catRadioElements.forEach(value => (value.checked = false));
+      this.updateTabIndex();
+    }
+  }
+
+  render() {
+    return (
+      <Host>
```
fynnfeldpausch, line 79: just one minor change here. I think you can remove <Host> element. It is only to be used, if you want to attach a class or something. If not, you can just directly return the div

fynnfeldpausch (approved):

PR #48, opened 2022-07-18
feat(core): updated tooltip ui

vardidzel:


core/src/components/cat-select/cat-select.tsx
```diff
@@ -149,7 +149,8 @@ export class CatSelect {
   }
 
   @Watch('value')
-  setValueHandler(value?: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
+  setValueHandler(value?: any) {
+    // eslint-disable-line @typescript-eslint/no-explicit-any
```
geralrodrigues, line 153: to pass the check of any (also prettier check), you can use eslint-disable-next-line above the method

core/src/components/cat-select/cat-select.tsx
```diff
@@ -149,12 +149,12 @@ export class CatSelect {
   }
 
   @Watch('value')
-  setValueHandler(value?: any) { // eslint-disable-line @typescript-eslint/no-explicit-any
+  setValueHandler(value?: never) {
```
fynnfeldpausch, line 152: ehh what? Why is the type `never`

core/src/components/cat-select/cat-select.tsx
```diff
@@ -149,12 +149,12 @@ export class CatSelect {
...
     if (this.resetItemsOnNextValueChange) {
       this.choice?.removeActiveItems(-1);
     }
     this.resetItemsOnNextValueChange = true;
-    this.choice?.setChoiceByValue(value);
+    this.choice?.setChoiceByValue(value || '');
```
fynnfeldpausch, line 157: I am also unsure about this change

fynnfeldpausch (changes requested):

fynnfeldpausch:
Prettier fails

vardidzel:
@fynnfeldpausch I have no idea how it was working before, but already fixed.

fynnfeldpausch (approved):

PR #49, opened 2022-07-19
feat(core): update notifications design

manyx:


core/src/components/cat-notification/cat-notification.tsx
```diff
@@ -123,15 +133,55 @@ class CatNotificationService {
...
+        x: '1.5rem',
+        y: '1.5rem'
+      }
+    };
+    Toastify(toastOptions).showToast();
+  }
+
+  primary(title: string, message = ' ', options?: Partial<ToastOptions>): void {
+    const position: ToastPosition = this.getPosition(options);
+    const toastOptions: Options = {
```
geralrodrigues, line 147: maybe you could put this option object in private method? to pass the principle of DRY, since it is the same in the other methods

manyx: done

geralrodrigues (approved):

PR #53, opened 2022-07-22
feat(core): upgrade alert component

geralrodrigues:


core/src/components/cat-alert/cat-alert.tsx
```diff
@@ -12,11 +12,29 @@ import { Component, h, Host, Prop } from '@stencil/core';
...
 
+  /**
+   * The name of an icon to be displayed in the alert.
+   */
+  @Prop() icon?: string;
+
+  /**
+   * Whether the icon of the alert  is deactivated.
+   */
+  @Prop() iconDeactivated = false;
```
fynnfeldpausch, line 36: Rename to noIcon

core/src/components/cat-alert/cat-alert.tsx
```diff
@@ -12,11 +12,29 @@ import { Component, h, Host, Prop } from '@stencil/core';
...
    */
   @Prop() color: 'primary' | 'secondary' | 'danger' | 'success' | 'warning' = 'primary';
 
+  /**
+   * The name of an icon to be displayed in the alert.
+   */
+  @Prop() icon?: string;
+
+  /**
+   * Whether the icon of the alert  is deactivated.
```
fynnfeldpausch, line 34: Remove double whitespace

core/src/components/cat-alert/cat-alert.scss
```diff
@@ -17,11 +17,16 @@
...
-  border-radius: cat-border-radius('m');
-  padding: 0.75rem 1rem;
+  border-radius: cat-border-radius('l');
+  padding: 1.25rem;
+  display: flex;
+  flex-direction: row;
+  gap: 0.5rem;
+}
+
+::slotted(*) {
```
fynnfeldpausch, line 28: Does this apply to all slotted content or only to direct children? I am always a bit cautious with * operator because of leaking styles everywhere

core/src/components/cat-alert/cat-alert.scss
```diff
@@ -33,11 +38,9 @@
...
   .alert-#{$theme} {
     --bg: #{cat-token('color.theme.#{$theme}.bg', $wrap: false)};
-    --text: #{cat-token('color.theme.#{$theme}.text', $wrap: false)};
-    @if $theme == 'secondary' {
-      --border: #{cat-token('color.theme.#{$theme}.bg', $wrap: false)};
-    } @else {
-      --border: #{cat-token('color.theme.#{$theme}.text', $wrap: false)};
+
+    cat-icon::part(icon) {
+      color: cat-token('color.theme.#{$theme}.bg');
```
fynnfeldpausch, line 43: This should not be theme.bg but theme.text, I think

core/src/index.html
```diff
@@ -79,6 +79,19 @@ <h3>Color</h3>
           <cat-alert color="danger">
             <p>Lorem ipsum dolor sit amet.</p>
           </cat-alert>
+          <h3>Variations</h3>
+          <cat-alert color="primary" icon-deactivated>
```
fynnfeldpausch, line 83: Keep in mind to update the option here after rename

fynnfeldpausch (changes requested):

core/src/components/cat-alert/cat-alert.scss
geralrodrigues: Only the direct children but still, I changed the selector because I noticed that it would be more complicated to add children

core/src/components/cat-alert/cat-alert.scss
geralrodrigues: certainly, but if i used text it wouldn't match with the design, so i added a new token to make it more readable

fynnfeldpausch (approved):

PR #54, opened 2022-07-22
feat(core): add input types

geralrodrigues:


core/src/components/cat-input/input-type.ts
```diff
@@ -0,0 +1,21 @@
+export type InputType =
+  | 'button'
+  | 'color'
+  | 'date'
+  | 'datetime-local'
+  | 'email'
+  | 'file'
+  | 'hidden'
```
fynnfeldpausch, line 8: I don't think hidden makes sense, as it will still render the web component. So I think we should remove that. Did you give the others a quick try to see if they are working?

fynnfeldpausch (changes requested):

core/src/components/cat-input/input-type.ts
geralrodrigues: Ok, yes, I did a quick check in the index with each one, there are some that look a bit strange but they all work, I will make a push with the different inputs in the index

fynnfeldpausch (approved):

PR #66, opened 2022-09-16
PMI-446: fix disabled select

manyx:


core/src/components/cat-select/cat-select.scss
```diff
@@ -60,6 +60,7 @@ label {
   background: cat-token('color.ui.background.inputDisabled');
   cursor: not-allowed;
   color: cat-token('color.ui.font.muted');
+  pointer-events: none;
```
fynnfeldpausch, line 63: even though this works, I still prefer a JS solution. We then stay consistent even if click is triggered programmatically. Please add

      if (this.disabled) {
        return;
      }

to the top of the `onClick` method

manyx: done

fynnfeldpausch (approved):

PR #67, opened 2022-10-04
Feat/tagging select

geralrodrigues:


core/src/components/cat-select/cat-select.tsx
```diff
@@ -65,6 +65,16 @@ export interface CatSelectState {
   totalElements?: number;
 }
 
+export interface CatSelectMultipleTaggingValue {
+  ids?: string[];
```
fynnfeldpausch, line 69: make these non-optional and use an empty array if it is empty. same for the single select tagging value

core/src/components/cat-select/cat-select.tsx
```diff
@@ -132,7 +142,7 @@ export class CatSelect {
   /**
    * The value of the select.
```
fynnfeldpausch, line 143: add some more docs here about the types and how they are influenced by the inputs

PR #72, opened 2022-10-24
fix(core): update disabled state

geralrodrigues:


core/src/components/cat-radio-group/cat-radio-group.tsx
```diff
@@ -67,9 +67,7 @@ export class CatRadioGroup {
   @Watch('disabled')
   onDisabledChanged(disabled: boolean) {
     this.catRadioGroup.forEach(catRadio => {
-      if (disabled) {
-        catRadio.disabled = disabled;
-      }
+      catRadio.disabled = disabled;
```
fynnfeldpausch, line 70: remove curly braces around statement

PR #74, opened 2022-10-26
fix(core): add disable overflow

geralrodrigues:


core/src/components/cat-menu/cat-menu.tsx
```diff
@@ -31,6 +31,11 @@ export class CatMenu {
    */
   @Prop() noAutoClose = false;
 
+  /**
+   * Disable the overflow when menu is open.
```
fynnfeldpausch, line 35: after rename, docs should be more like "Allow overflow when menu is open."

fynnfeldpausch (approved):

PR #96, opened 2022-11-22
fix(core): fixed bug

manyx:
Fix badge not updating on property changes


core/src/components/cat-badge/cat-badge.tsx
```diff
@@ -47,4 +47,10 @@ export class CatBadge {
   render() {
     return <slot></slot>;
   }
+
+  componentWillRender(): void {
```
geralrodrigues, line 51: Put the method above the render

manyx: done

geralrodrigues (approved):

PR #101, opened 2022-12-01
Fix/select

geralrodrigues:


core/src/components/cat-select/cat-select.tsx
```diff
@@ -21,6 +21,7 @@ import {
 } from 'rxjs';
 import { CatFormHint } from '../cat-form-hint/cat-form-hint';
 import { catI18nRegistry as i18n } from '../cat-i18n/cat-i18n-registry';
+import isEqual from 'lodash/isEqual';
```
fynnfeldpausch, line 24: I would really like to keep this project lodash free. Is there any option to implement this without lodash. We are trying to get rid of lodash in the core project for various reasons and my fear is that once it is introduced here, we will use it more and more in this project

geralrodrigues: Hmm okey, I will remove it

core/src/components/cat-select/cat-select.tsx
```diff
@@ -119,6 +119,7 @@ export class CatSelect {
   private subscription?: Subscription;
   private term$: Subject<string> = new Subject();
   private more$: Subject<void> = new Subject();
+  private isUserChanged = false;
```
fynnfeldpausch, line 122: @geralrodrigues Why is this variable called isUserChanged? it could be any value, right?

geralrodrigues: Only boolean, it's a flag that I am using to know if it is a change by the UI or programmatically. Basically, if it is programmatic, it goes through value first and then selection, and through the ui, first selection and then value

geralrodrigues: Should a better name be used? Now I put the type too

geralrodrigues: Well, I cannot put the type😅 
image

fynnfeldpausch (approved):

PR #107, opened 2022-12-21
Feat/form group

geralrodrigues:


core/src/components/cat-input/cat-input.tsx
```diff
@@ -210,11 +215,16 @@ export class CatInput {
           <label htmlFor={this.id} class={{ hidden: this.labelHidden }}>
             <span part="label">
               {(this.hasSlottedLabel && <slot name="label"></slot>) || this.label}
-              {!this.required && (
+              {!this.required && (!this.requiredMarker || this.requiredMarker === 'optional') && (
```
fynnfeldpausch, line 218: remove the first part of the condition and set a default value for the property
@Prop() requiredMarker: 'none' | 'required' | 'optional' = 'optional';

fynnfeldpausch: same for the other form elements

core/src/components/cat-input/cat-input.tsx
```diff
@@ -210,11 +215,16 @@ export class CatInput {
...
                 <span class="input-optional" aria-hidden="true">
                   ({i18n.t('input.optional')})
                 </span>
               )}
+              {this.required && (!this.requiredMarker || this.requiredMarker === 'required') && (
```
fynnfeldpausch, line 223: also simplify this after setting a default value

core/src/components/cat-form-group/cat-form-group.tsx
```diff
@@ -0,0 +1,62 @@
+import { Component, Element, h, Host, Prop, Watch } from '@stencil/core';
+
+type HTMLCatFormFieldElement =
+  | HTMLCatInputElement
+  | HTMLCatTextareaElement
+  | HTMLCatSelectElement
+  | HTMLCatLabelElement;
```
fynnfeldpausch, line 7: add newline here

core/src/components/cat-form-group/cat-form-group.tsx
```diff
@@ -0,0 +1,62 @@
...
+@Component({
+  tag: 'cat-form-group',
+  styleUrl: 'cat-form-group.scss',
+  shadow: true
+})
+export class CatFormGroup {
+  private formElements?: HTMLCatFormFieldElement[];
```
fynnfeldpausch, line 14: can we not just make this non-optional and assign an empty array here?

core/src/components/cat-form-group/cat-form-group.tsx
```diff
@@ -0,0 +1,62 @@
...
+   *
+   * If there are more required, the optional will be marked.
+   * If there are less required, it will mark the required.
+   */
+  @Prop() requiredMarker: 'none' | 'required' | 'optional' | 'auto' = 'auto';
+
+  @Watch('requiredMarker')
+  onRequiredMarker(newRequiredMarker: 'none' | 'required' | 'optional' | 'auto') {
+    let updateMarker;
+    if (this.formElements) {
```
fynnfeldpausch, line 31: if it has a default value, I guess this can be removed

core/src/components/cat-form-group/cat-form-group.tsx
```diff
@@ -0,0 +1,62 @@
...
+      for (const element of this.formElements) {
+        if (!element.requiredMarker) {
+          element.requiredMarker = updateMarker;
+        }
+      }
+    }
+  }
+
+  componentDidLoad(): void {
+    this.formElements = Array.from(
```
fynnfeldpausch, line 42: for now, this is ok, but I guess we need to optimize this at some point because often some form elements are dynamically added using ngIf. Then this breaks

core/src/components/cat-form-group/cat-form-group.tsx
```diff
@@ -0,0 +1,62 @@
...
+  @Prop() requiredMarker: 'none' | 'required' | 'optional' | 'auto' = 'auto';
+
+  @Watch('requiredMarker')
+  onRequiredMarker(newRequiredMarker: 'none' | 'required' | 'optional' | 'auto') {
+    let updateMarker;
+    if (this.formElements) {
+      updateMarker = newRequiredMarker === 'auto' ? this.calculate(this.formElements) : newRequiredMarker;
+      for (const element of this.formElements) {
+        if (!element.requiredMarker) {
+          element.requiredMarker = updateMarker;
```
fynnfeldpausch, line 35: will this lead to a problem where the change only works once? because the initial change, the requiredMarker is not set, but then it is set by the parent component. If then the requiredMarker changes in the parent, it will skip this component, because it has already been set. Know what I mean? Don't know a good way how to fix it thoguh

fynnfeldpausch (changes requested):

core/src/components/cat-form-group/cat-form-group.tsx
geralrodrigues: hmm yes, I understand, I'll try to think of a solution too

geralrodrigues: At the moment if we put the required marker with a default value in each element, my first solution would not work.

A solution that could work, could be to add more options or a variable in the child elements to mark that it cannot be overridden, and that the parent can always modify that variable, except if it is an option or has that mark. For example, if the `requireMarked` has “!important”

Or that the parent can always overwrite the variable without restrictions, and thus only have a source of truth, in the case that a child wants to overwrite that variable, then the structure is changed, that is, that it is not wrapped by the parent anymore.

fynnfeldpausch (approved):

PR #131, opened 2023-01-26
feat(core): fix avatar initials in select component

emmoldovan:


core/src/components/cat-select/cat-select.tsx
```diff
@@ -577,7 +578,8 @@ export class CatSelect {
                 label={this.state.selection[0].render.label}
                 round={this.state.selection[0].render.avatar.round}
                 src={this.state.selection[0].render.avatar.src}
-                initials={''}
+                initials={this.state.selection[0].render.avatar.initials ?? ''}
+                size={this.state.selection[0].render.avatar.initials ? 'xs' : 'm'}
```
fynnfeldpausch, line 582: connecting the size of the avatar to the initials does not make sense to me. Why was it implemented this way?

fynnfeldpausch (changes requested):

core/src/components/cat-avatar/cat-avatar.scss
```diff
@@ -43,6 +43,7 @@ $-sizes: (
   white-space: nowrap;
   overflow: hidden;
   vertical-align: middle;
+  font-weight: 600;
```
geralrodrigues, line 46: It is better to modify the font-weight from the style sheet of the select, since we only want to modify the initials inside the select, right?

fynnfeldpausch: Why do we want to change the font weight in the first place? The avatar should be an independent component and should not be influenced by this. What is the reasoning @mrkskmn ?

mrkskmn: I am missing context. 
Is this about https://haiilo.atlassian.net/browse/PMI-819? 
If so, I was just handing off the text styles from Figma. Maybe there was a mismatch between the code and Figma if they haven't had the font-weight:600 already.

fynnfeldpausch: ok, then this works 👍

PR #132, opened 2023-01-26
feat(core): form-alignment

fynnfeldpausch:


core/src/components/cat-input/cat-input.scss
```diff
@@ -4,9 +4,8 @@
 @use 'utils/color';
 
 :host {
+  --input-height: 2.5rem;
```
fynnfeldpausch, line 7: Why has this become a variable?

emmoldovan: Because the min height of the label depends on the input height

fynnfeldpausch: but the input height cannot be changed, right? If it is just about reusing the value, than we should use a private SCSS variable here. And also in the textarea. My question was, why is it a CSS custom property and not a SCSS variable.

emmoldovan: Yes, you are right. It should have been a scss variable

fynnfeldpausch: ok, please correct this (also in textarea, i guess) and then we can merge

PR #147, opened 2023-02-10
Feature/character limit

emmoldovan:


core/src/styles/core/_form.scss
```diff
@@ -2,7 +2,67 @@
 @use '../mixins' as *;
```
fynnfeldpausch, line 2: I know, I said that we should move this into a separate SCSS file. But I changed my mind 🤣 

The reason is that web components are designed to be self contained components. And now we are very much relying on outside styles, which is something we want to avoid. So lets create a new folder at `styles/_snippets` and move the code there in a file called `_form-label`. This file should not be included in the global styles. Instead, we can load it in both the input and textarea styles. Thus we make it reusable but still have the components own the styles.

core/src/styles/core/_form.scss
```diff
@@ -2,7 +2,67 @@
...
 
 // ---- Form
+$cat-input-height: 2.5rem;
 
 .cat-error {
```
fynnfeldpausch, line 7: as this is also deprecated, we can remove the _form file altogether after you have moved the code.

fynnfeldpausch (changes requested):

PR #178, opened 2023-03-13
feat(core): create new button-group component

emmoldovan:


core/src/components/cat-button-group/cat-button-group.tsx
```diff
@@ -0,0 +1,41 @@
...
+  render() {
+    return (
+      <Host role="group" aria-label={this.a11yLabel}>
+        <slot></slot>
+      </Host>
+    );
+  }
+
+  componentDidLoad() {
+    this.formElements = Array.from(this.hostElement.querySelectorAll('cat-button')) as HTMLCatButtonElement[];
```
fynnfeldpausch, line 35: I think the typecast at the end can be deleted

core/src/components/cat-button-group/cat-button-group.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    this.formElements.forEach((element, index) => {
+      element.buttonGroupPosition = index === 0 ? 'first' : index === this.formElements.length - 1 ? 'last' : 'middle';
```
fynnfeldpausch, line 38: not sure if this is a problem but this will only be executed once at the beginning. This means if we have a button group in angular with an ngIf on some buttons, the position will be wrong if the dynamically show up. Do you think we should execute this on ever componentDidRender?

core/src/components/cat-button/cat-button.scss
```diff
@@ -78,6 +78,29 @@ $button-sizes: (
...
+  }
+
+  &-middle {
+    border-radius: 0;
+  }
+
+  &-last {
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+    border-right: none;
```
fynnfeldpausch, line 96: im confused. why does the last button have no right border?

fynnfeldpausch (changes requested):

core/src/components/cat-button-group/cat-button-group.scss
```diff
@@ -0,0 +1,20 @@
...
+  display: inline-flex;
+  vertical-align: middle;
+}
+
+::slotted(cat-button[variant='outlined']),
+::slotted(cat-button:not([variant])) {
+  margin-right: -1px;
+}
+
+::slotted(cat-button[variant='outlined']:first-child),
```
fynnfeldpausch, line 12: why is this needed?

core/src/components/cat-button-group/cat-button-group.tsx
fynnfeldpausch: wow, didnt know this is an option.

core/src/components/cat-button/cat-button.scss
```diff
@@ -78,6 +78,32 @@ $button-sizes: (
...
+  &-middle {
+    border-radius: 0;
+  }
+
+  &-last {
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+  }
+
+  &:focus-visible {
```
fynnfeldpausch, line 98: shouldn't this be after the &:hover rule? otherwise if there is focus and we hover, it will be back to 1

fynnfeldpausch (changes requested):

core/src/components/cat-button-group/cat-button-group.scss
emmoldovan: Yeap, leftovers

fynnfeldpausch (approved):

PR #194, opened 2023-03-22
fix(core): Change input fields box-shadow to inset to match cat-button inset box-shadow

emmoldovan:


fynnfeldpausch:
This is outdated, right @emmoldovan ?

PR #203, opened 2023-04-12
Feature/cat datpicker component

emmoldovan:


core/package.json
```diff
@@ -50,10 +50,12 @@
     "@stencil/core": "3.0.0",
     "autosize": "6.0.0",
     "autosize-input": "1.0.2",
+    "dayjs": "^1.11.7",
     "focus-trap": "7.2.0",
     "loglevel": "1.8.1",
     "sanitize.css": "13.0.0",
-    "toastify-js": "1.12.0"
+    "toastify-js": "1.12.0",
+    "vanillajs-datepicker": "^1.3.1"
```
fynnfeldpausch, line 58: pin version

core/package.json
```diff
@@ -50,10 +50,12 @@
     "@stencil/core": "3.0.0",
     "autosize": "6.0.0",
     "autosize-input": "1.0.2",
+    "dayjs": "^1.11.7",
```
fynnfeldpausch, line 53: can we somehow remove this dependency?

fynnfeldpausch:
General observations:
* day view: hover not working for current day
* SCSS vars need to be mapped to catalyst variables, we can do that together at some point
* week picker: current week should be highlighted, not current day (if possible)
* week picker: hover style should not extra-highlight first day of week
* month picker: can / should we highlight the current month?
* month picker: the months being higher than wide looks odd to me
* year view: see both points for month view as well

fynnfeldpausch (changes requested):

PR #243, opened 2023-06-08
fix(core): add output event to get page

Jeff1899:
Add output event to get page number, when component changed

### DIFF
--- a/core/src/components/cat-pagination/cat-pagination.tsx
+++ b/core/src/components/cat-pagination/cat-pagination.tsx
@@ -86,7 +86,7 @@ export class CatPagination {
               a11yLabel={i18n.t('pagination.prev')}
               icon={this.iconPrev}
               iconOnly
-              onClick={() => this.prevPage()}
+              onClick={() => this.setPage(this.page - 1)}
             ></cat-button>
           </li>
           {this.content}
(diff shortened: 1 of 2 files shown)


core/src/components/cat-pagination/cat-pagination.tsx
```diff
@@ -81,7 +86,7 @@ export class CatPagination {
               a11yLabel={i18n.t('pagination.prev')}
               icon={this.iconPrev}
               iconOnly
-              onClick={() => (this.page = this.page - 1)}
+              onClick={() => this.prevPage()}
```
fynnfeldpausch, line 89: as this has not been checked before either, I would just rely on the fact, that the button will be disabled. This just, call setPage here. Same goes for nextPage.

core/src/components/cat-pagination/cat-pagination.tsx
```diff
@@ -110,6 +115,23 @@ export class CatPagination {
     return this.page === this.pageCount - 1;
   }
 
+  setPage(value: number) {
```
fynnfeldpausch, line 118: im not 100% sure, but you should be able to make this private.

fynnfeldpausch:
Nice work. Please also add a console output to the index.html as stated in the ticket <3

fynnfeldpausch (changes requested):

Jeff1899:
> Nice work. Please also add a console output to the index.html as stated in the ticket <3

Ahh I thought that was just a tip to test it myself, thats why I removed it afterwards.
I hope it fits now, I'm not sure how you have imagined that

fynnfeldpausch (approved):

PR #269, opened 2023-06-14
fix(core): stop event propagation on cat-select open/close arrow and …

emmoldovan:
fix(core): stop event propagation on cat-select open/close arrow and  clear button


core/src/components/cat-select/cat-select.tsx
```diff
@@ -615,6 +615,7 @@ export class CatSelect {
                             a11yLabel={i18n.t('select.deselect')}
                             onClick={() => this.deselect(item.item.id)}
                             tabIndex={-1}
+                            onCatClick={event => event.stopImmediatePropagation()}
```
fynnfeldpausch, line 618: check stopPropagation() and change if possible

PR #278, opened 2023-06-23
Feat/formly integration

emmoldovan:


angular/projects/catalyst-demo/src/app/app.component.html
```diff
@@ -4,8 +4,10 @@ <h2>Dialog</h2>
 <h2>Form Components</h2>
 <form [formGroup]="form">
   <cat-input label="Input" formControlName="test" textPrefix="write 'aaa' to make it valid" [nativeAttributes]="{'data-test': 'test-attribute'}" hint="this is a hint"></cat-input>
-  <cat-input label="relatedInput" formControlName="relatedInput" [errorUpdate]="4000"></cat-input>
+  <cat-input label="relatedInput" formControlName="relatedInput"></cat-input>
```
fynnfeldpausch, line 7: Is there a reason for this change related to formly? Otherwise, revert.

angular/projects/catalyst-demo/src/app/app.component.html
```diff
@@ -4,8 +4,10 @@ <h2>Dialog</h2>
...
   <cat-textarea label="Textarea" formControlName="test" [nativeAttributes]="{'data-test': 'test-attribute'}" hint="this is a hint"></cat-textarea>
+
```
fynnfeldpausch, line 9: Is there a reason for this change related to formly? Otherwise, revert.

angular/projects/catalyst-demo/src/app/app.component.ts
```diff
@@ -29,7 +31,7 @@ const countries: Country[] = [
 export class AppComponent implements OnInit {
   form = new FormGroup({
     test: new FormControl('test', [Validators.pattern('a+'), Validators.required, Validators.minLength(3)]),
-    relatedInput: new FormControl(null, [this.equalTo('test')]),
+    relatedInput: new FormControl(null, [Validators.required]),
```
fynnfeldpausch, line 34: Is there a reason for this change related to formly? Otherwise, revert.

angular/projects/catalyst-demo/src/app/app.component.ts
```diff
@@ -40,6 +42,93 @@ export class AppComponent implements OnInit {
...
+            value: 'three',
+            checked: true,
+          },
+        ]
+      }
+    },
+    {
+      key: 'catDatepicker',
+      type: CatDatepickerFieldType,
+      defaultValue: '12.12.2022',
```
fynnfeldpausch, line 113: does this work? I thought I built it in a way where the date is expected as ISO string. I'd change it to an ISO string

angular/projects/catalyst-demo/src/app/app.component.ts
```diff
@@ -40,6 +42,93 @@ export class AppComponent implements OnInit {
...
+        mode: 'datetime',
+        required: true
+      },
+    },
+    {
+      key: 'catSelect',
+      type: CatSelectFieldType,      
+      props: {
+        label: 'Select a country',
+        connector: this.countryConnector,
```
fynnfeldpausch, line 126: nice!

angular/projects/catalyst/package.json
```diff
@@ -12,7 +12,8 @@
   "peerDependencies": {
     "@angular/core": ">=14.0.0",
     "@angular/cdk": ">=14.0.0",
-    "rxjs": ">=6.6.0"
+    "rxjs": ">=6.6.0",
+    "@ngx-formly/core": ">=6.1.8"
```
fynnfeldpausch, line 16: Any idea how to resolve this? We now force the user to install ngx-formly because otherwise there is a dependency issue. Ideally it would be an optional dependency, but I have no idea ho that works. Because you still need it to prevent compile errors

angular/projects/catalyst/src/lib/formly/formly-checkbox/formly-checkbox.component.ts
```diff
@@ -0,0 +1,22 @@
...
+      [label]="props.label"
+      [name]="props.name"
+      [hint]="props.hint"
+      [labelLeft]="props.labelLeft"
+      [nativeAttributes]="props.nativeAttributes"
+      >
+    </cat-checkbox>
+  `,
+})
+export class CatCheckboxFieldType extends FieldType<FieldTypeConfig> {}
```
fynnfeldpausch, line 22: please add newlines to each file ending. This is our standard and should lead to a lint error 🤔

angular/projects/catalyst/src/lib/formly/formly-select/formly-select.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    </cat-select>
+  `,
+})
+export class CatSelectFieldType extends FieldType<FieldTypeConfig> implements AfterViewInit {
+  @ViewChild('selectRef') selectRef!: CatSelect;
+  
+  ngAfterViewInit(): void {
+    this.selectRef.connect(this.props.connector)
+  }
+
```
fynnfeldpausch, line 40: remove these newlines

angular/projects/catalyst/src/lib/formly/formly-select/formly-select.component.ts
```diff
@@ -0,0 +1,42 @@
...
+      #selectRef
+      >
+    </cat-select>
+  `,
+})
+export class CatSelectFieldType extends FieldType<FieldTypeConfig> implements AfterViewInit {
+  @ViewChild('selectRef') selectRef!: CatSelect;
+  
+  ngAfterViewInit(): void {
+    this.selectRef.connect(this.props.connector)
```
fynnfeldpausch, line 38: is the template using the angular wrapper or the underlying stencil web component? Because the angular wrapper also provides a connector property binding I think. I built that in angular/projects/catalyst/src/lib/directives/select-value-accessor-decorator.ts

angular/projects/catalyst/src/lib/formly/formly.module.ts
```diff
@@ -0,0 +1,78 @@
...
+    CatToggleFieldType,
+    CatRadioFieldType,
+    CatRadioGroupFieldType,
+    CatInputFieldType,
+    CatTextareaFieldType,
+    CatDatepickerFieldType,
+    CatSelectFieldType
+  ],
+})
+export class CatalystFormlyModule {}
```
fynnfeldpausch, line 78: Why are we not importing this anywhere in the demo app? Shouldn't we have to?

fynnfeldpausch (changes requested):
Nice and clean solution. Great!

angular/projects/catalyst-demo/src/app/app.component.ts
emmoldovan: Yes, it works this way

angular/projects/catalyst/package.json
emmoldovan: propably a solution is to split the library smaller libraries like. catalyst/components, catalyst/forms etc

angular/projects/catalyst/src/lib/formly/formly.module.ts
emmoldovan: I've exported also each component separately in the public-api file.. Probably thats why the components are imported directly from the catalyst library. But if you want to use the custom aliases you should import the CatalystFormlyModule
image

angular/projects/catalyst/src/lib/formly/formly-select/formly-select.component.ts
emmoldovan: Do you mean this one?
image

fynnfeldpausch (approved):

Issue #151, opened 2023-02-14
Cat select dropdown not positioned correctly


fynnfeldpausch:

emmoldovan:
#288

PR #288, opened 2023-07-09
fix(core): enable cat-select dropdown flip if there is available space

emmoldovan:


emmoldovan:
#151

core/src/components/cat-select/cat-select.tsx
```diff
@@ -118,7 +118,7 @@ export class CatSelect {
     return this.identifier || this._id;
   }
 
-  private dropdown?: HTMLElement;
+  private dropdown!: HTMLElement;
```
fynnfeldpausch, line 121: try to avoid this change. It is technically correct to use the ? here

core/src/components/cat-select/cat-select.tsx
```diff
@@ -696,7 +696,7 @@ export class CatSelect {
 
         <div
           class="select-dropdown"
-          ref={el => (this.dropdown = el)}
+          ref={el => (this.dropdown = el as HTMLElement)}
```
fynnfeldpausch, line 699: revert

core/src/components/cat-select/cat-select.tsx
```diff
@@ -973,11 +973,14 @@ export class CatSelect {
   private update() {
     if (this.trigger && this.dropdown) {
       computePosition(this.trigger, this.dropdown, {
+        strategy: 'fixed',
         placement: this.placement,
-        middleware: [offset(CatSelect.DROPDOWN_OFFSET)]
-      }).then(({ x, y }) => {
+        middleware: [offset(CatSelect.DROPDOWN_OFFSET), flip()]
+      }).then(({ x, y, placement }) => {
+        this.dropdown.dataset.placement = placement;
```
fynnfeldpausch, line 980: if you move this into the if condition, you should be fine with the conditional

fynnfeldpausch (changes requested):

PR #293, opened 2023-07-11
feat(core): adjust tabs padding and hover & active states

### LINKED ISSUE #199: Adjust tab component to be the same as tabs in Studio
We re-styled the nav links inside the Studios MFE navigation to be like tabs. We want to use the same styling for the real tab component. 

@fynnfeldpausch did the re-styled tab implementation. There's also a nice little trick in there to avoid UI flickering when applying a thicker font-weigt to the active tab ([see this thread ->](https://stackoverflow.com/questions/556153/inline-elements-shifting-when-made-bold-on-hover))

CleanShot 2023-04-04 at 09 38 07@2x

emmoldovan:
closes #199


emmoldovan:
I saw that on the Tab/Tabs component we do not have proper theming implemented. I implemented without theming in mind for now

fynnfeldpausch:
@emmoldovan What kind of theming is missing?

fynnfeldpausch (approved):

fynnfeldpausch:
Looks good from my side, but @mrkskmn can you also have another look if this looks like expected?

mrkskmn:
**I noticed 2 issues**

1. Tabs cannot be toggled when clicking the padding. This makes the click area too small and frustrating to use. Can we change it so we can make the padding clickable as well? 
2. The fix with the :before pseudo-element to avoid flickering when switching to active does not work when an icon is set. Can we fix that? 

https://github.com/haiilo/catalyst/assets/98806154/94c0f348-8f83-46f4-be40-94ae3bd57766

core/src/components/cat-button/cat-button.scss
mrkskmn, line 1: The padding of "1.125rem 0.75rem" should be applied here to the button instead of the cat-tab class

core/src/components/cat-tabs/cat-tabs.scss
```diff
@@ -26,9 +26,47 @@
...
 
 .cat-tab {
-  padding: 0.5rem 0;
+  display: inline-flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: space-between;
+  position: relative;
+  transition: none;
+  padding: 1.125rem 0.75rem;
```
mrkskmn, line 35: The padding of "1.125rem 0.75rem" should be removed here and be applied to the cat-button instead

mrkskmn (changes requested):
Found the issue about the click area

core/src/components/cat-tabs/cat-tabs.scss
fynnfeldpausch: fixed

fynnfeldpausch:
> 2. The fix with the ":before" pseudo-element to avoid flickering when switching to active does not work when an icon is set. Can we fix that?

Hmm.. don't really find a way to fix it. will look into it

fynnfeldpausch:
@mrkskmn Please have another look. I think I found a solution for the layout shift. That was a tricky one. Also have a quick look at buttons, to see if everything still looks like expected, as I changed something there as well. But to me it looked good.

emmoldovan (commented):
Nice approach

PR #300, opened 2023-07-17
21109: fix radio group listening to non radio events

finnfabians:


finnfabians:
Radio group value and selection will not change

kanstantsin-malikau (approved):

PR #305, opened 2023-07-18
feat(core): dropdown improvements

mrkskmn:
**Dropdown improvements**
* Re-added a consistent padding to dropdown body (affects inner buttons)
* Changed transition from a "slide fade in" to a "scale fade in"
* Fixed an issue with the transition where the first time opening a dropdown would not trigger the transition
* Added new real life example to stencil page

### DIFF
--- a/core/src/components/cat-dropdown/cat-dropdown.scss
+++ b/core/src/components/cat-dropdown/cat-dropdown.scss
@@ -15,7 +15,7 @@ $-max-height: calc(100vh - 48px);
 }
 
 ::slotted(nav) {
-  padding: cat-token('size.spacing.xs');
+  padding: 0.25rem;
   min-width: $-min-width;
   max-width: $-max-width;
 }
--- a/core/src/components/cat-dropdown/cat-dropdown.scss
+++ b/core/src/components/cat-dropdown/cat-dropdown.scss
@@ -15,8 +15,7 @@ $-max-height: calc(100vh - 48px);
 }
 
 ::slotted(nav) {
-  padding-top: $cat-body-margin-bottom * 0.5;
-  padding-bottom: $cat-body-margin-bottom * 0.5;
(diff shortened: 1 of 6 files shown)


core/src/components/cat-dropdown/cat-dropdown.scss
```diff
@@ -15,49 +15,83 @@ $-max-height: calc(100vh - 48px);
 }
 
 ::slotted(nav) {
-  padding-top: $cat-body-margin-bottom * 0.5;
-  padding-bottom: $cat-body-margin-bottom * 0.5;
+  padding: cat-token('size.spacing.xs');
```
fynnfeldpausch, line 18: spacing values are deprecated as tokens. they only exist to generate the correct spacing classes. Use hard-coded spacings here.

core/src/styles/core/_nav.scss
```diff
@@ -4,13 +4,12 @@
 nav {
   hr,
   .cat-hr {
-    margin-top: $cat-nav-padding-horizontal * 0.5;
-    margin-bottom: $cat-nav-padding-horizontal * 0.5;
+    margin: cat-token('size.spacing.xs') calc(-1 * $cat-nav-padding-horizontal);
```
fynnfeldpausch, line 7: spacing values are deprecated as tokens. they only exist to generate the correct spacing classes. Use hard-coded spacings here.

fynnfeldpausch (changes requested):

core/src/components/cat-dropdown/cat-dropdown.scss
mrkskmn: Oh I was not aware of that. Do we not want to have spacing tokens anymore?

fynnfeldpausch (approved):

PR #309, opened 2023-07-18
fix(core): update datepicker while handling via form

Jeff1899:


core/src/components/cat-datepicker/cat-datepicker.tsx
```diff
@@ -5,6 +5,7 @@ import { ErrorMap } from '../cat-form-hint/cat-form-hint';
 import { catI18nRegistry as i18n } from '../cat-i18n/cat-i18n-registry';
 import { getFormat } from './cat-datepicker.config';
 import { getLocale } from './cat-datepicker.locale';
+import { Watch } from '../../../dist/types/stencil-public-runtime';
```
fynnfeldpausch, line 8: This seems to be the wrong import path

core/src/components/cat-datepicker/cat-datepicker.tsx
```diff
@@ -249,9 +264,8 @@ export class CatDatepickerFlat {
         onCatChange={e => {
           e.stopPropagation();
           this.pickr?.setDate(e.detail);
```
fynnfeldpausch, line 266: if we are setting the value below and the watch kicks in, shouldnt this line be redundant and be deleted?

core/src/components/cat-datepicker/cat-datepicker.tsx
```diff
@@ -249,9 +264,8 @@ export class CatDatepickerFlat {
...
-          if (this.value !== (e.detail || undefined)) {
+          if (this.value !== (e.detail || undefined || '')) {
```
fynnfeldpausch, line 267: why do we need || '' here?

fynnfeldpausch (changes requested):

fynnfeldpausch (approved):

hjalmararnorssonhaiilo:
Hi, I had a look at this fix in the latest Catalyst version, 6.2.3. It seems that this issue is still there as programmatic changes to the bound formcontrol value does not update the selected date in the datepicker UI

PR #327, opened 2023-07-26
feat: minor follow up improvements for tabs

mrkskmn:
Adjusted spacings for tabs which align to an edge (first or last).


fynnfeldpausch:
This still seems buggy to me. justified tabs should not have a space between them IMO

fynnfeldpausch (changes requested):
Lets fix justified tabs first

fynnfeldpausch (approved):

PR #334, opened 2023-07-28
feat(core): Added source sans 3 as variable default font

mrkskmn:


mrkskmn:
This leads to the issue with vertical metrics again which we previously "solved" by hacking the lato font and hosting it on our own. :/

mrkskmn:
> This leads to the issue with vertical metrics again which we previously "solved" by hacking the lato font and hosting it on our own. :/

To fix the most obvious issue with this - inside buttons - a fix would be to set the line-height fix to 1.1 to all buttons. For some reason it is fixed when using odd line-height values.

fynnfeldpausch:
Any clue what the true cause is and how we can fix it without hacking?

mrkskmn:
Normally yes, that issue is usually caused by misaligned vertical metrics within the font file. Ideally bot top and bottom spacings of a fonts metrics are set to the same values. If that's not the case, the font will render differently/misaligned on different platforms. This article by Google Fonts describes this very good: https://fonts.google.com/knowledge/using_type/vertical_spacing_and_line_height_in_design_systems

Source Sans Pro however already has equally set vertical metrics for top and bottom spacings. I also noticed in my implementation that on some parts the issues does not occur (e.g. regular p text) but in some components, such as the button, it is vertically misaligned by roughly 1-2 pixels, resulting in an optical shift. Very strange. 

But then again it is just me staring too long at this "problem". Did the labels seem misaligned to you inside the button component?

fynnfeldpausch (approved):

fynnfeldpausch (approved):

PR #360, opened 2023-08-16
chore(core): fixed broken spacings in nav cat-card

mrkskmn:
I basically just removed the conditional styling for nav elements inside cat-cards as we do not need that anymore.
This would also mean that a card that's supposed to contain a nav element also needs the class "cat-p-xs" to work properly. Is that an acceptable solution or too manual?


fynnfeldpausch:
too manual :D

fynnfeldpausch (approved):

PR #361, opened 2023-08-16
feat: new type section on overview page

mrkskmn:
- New type section
- Smaller adjustments to the top buttons of the page (github, slack)
- Added icon sprite page as a 3rd tab
- Added switch buttons for font and theme
- Initial proposal for typography spacings inside <article> tag

### DIFF
--- a/core/src/styles/core/_typography.scss
+++ b/core/src/styles/core/_typography.scss
@@ -241,11 +241,11 @@ code {
 
 // ----- editorial typography
 
-article{
-
+cat-article,
+.cat-article {
   font-size: 1rem;
 
-  $cat-head-margin-bottom: 1.5rem; 
+  $cat-head-margin-bottom: 1.5rem;
   $cat-body-margin-bottom: 2.5rem;
 
   :where(h1, h2, h3, h4, h5, h6),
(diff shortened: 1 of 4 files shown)


core/src/styles/core/_typography.scss
```diff
@@ -238,3 +238,62 @@ code {
...
+  .cat-h2,
+  .cat-h3,
+  .cat-h4,
+  .cat-h5,
+  .cat-h6 {
+    margin-top: 0;
+    margin-bottom: $cat-head-margin-bottom;
+  }
+
+  :where(h1, h2, h3, h4, h5, h6, .cat-h1, .cat-h2, .cat-h3, .cat-h4, .cat-h5, .cat-h6):has(+ h1, + h2, + h3, + h4, + h5, + h6, + .cat-h1, + .cat-h2, + .cat-h3, + .cat-h4, + .cat-h5, + .cat-h6 ) {
```
fynnfeldpausch, line 262: we cannot rely on has without Firefox support :(

mrkskmn: I know, unfortunately :(. 
Currently looking for a fallback solution for firefox. Our current % of Firefox users for Haiilo is around 4.5%. 

Do you know about any proper fallback?

mrkskmn: If we wait long enough we might just get Firefox support. 😅  
The are currently working on it and it is supposed to be launched in 2023
https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idi-p/23794/page/3#comments

core/src/styles/core/_typography.scss
```diff
@@ -238,3 +238,62 @@ code {
   @include cat-break-word;
   padding: 0.0625rem 0.25rem;
 }
+
+// ----- editorial typography
+
+article{
```
fynnfeldpausch, line 244: change to a specific class

mrkskmn:
Added in what we talked about @fynnfeldpausch. 

I also replaced the :has() selector with a prototyped JS fallback which essentially does the same thing as the CSS selector. Works nicely in all major browsers for our demo page. However, as this script is executed once I am unsure about the real-world use in e.g. Intranet, where content can appear dynamically. Could such a JS solution imply performance issues?

fynnfeldpausch (approved):
ok as a start

PR #391, opened 2023-10-05
chore(core): [a11y] error handling for inputs

malikov917:
more details why: https://haiilo.atlassian.net/browse/COYOFOUR-21110

### DIFF
--- a/core/src/components/cat-form-hint/cat-form-hint.tsx
+++ b/core/src/components/cat-form-hint/cat-form-hint.tsx
@@ -28,10 +28,10 @@ export const CatFormHint: FunctionalComponent<CatFormHintProps> = props => {
   const { id, hint, slottedHint, errorMap } = props;
   const errors = Object.entries(errorMap || {});
   return (
-    <div id={id + '-hint'} class="hint-section">
+    <div aria-live="polite" id={id + '-hint'} class="hint-section">
       {errors.length
         ? errors.map(([key, params]) => (
-            <p tabindex="0" role="alert" class={(props.class ?? '') + ' input-hint cat-text-danger'}>{i18n.t(`error.${key}`, params)}</p>
+            <p class={(props.class ?? '') + ' input-hint cat-text-danger'}>{i18n.t(`error.${key}`, params)}</p>
           ))
         : [
             hint &&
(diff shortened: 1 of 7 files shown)


core/src/components/cat-form-hint/cat-form-hint.tsx
```diff
@@ -31,7 +31,7 @@ export const CatFormHint: FunctionalComponent<CatFormHintProps> = props => {
     <div id={id + '-hint'} class="hint-section">
       {errors.length
         ? errors.map(([key, params]) => (
-            <p class={(props.class ?? '') + ' input-hint cat-text-danger'}>{i18n.t(`error.${key}`, params)}</p>
+            <p tabindex="0" role="alert" class={(props.class ?? '') + ' input-hint cat-text-danger'}>{i18n.t(`error.${key}`, params)}</p>
```
TimSielemann, line 34: this is not an interactive element so it should not be tabable, also i think role=alert does not work with this dynamically created element. aria-live is only triggered for an element that changes it's content not on new elements (see hint box here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) 

Additionally this does not solve the described by issue that is only added to the input when there is a hint but should also be added when there is an error.

kanstantsin-malikau: your comment is resolved, now input errors support live updates in voice over

core/src/components/cat-input/cat-input.tsx
```diff
@@ -325,7 +325,7 @@ export class CatInput {
                 onFocus={this.onFocus.bind(this)}
                 onBlur={this.onBlur.bind(this)}
                 aria-invalid={this.invalid ? 'true' : undefined}
-                aria-describedby={this.hint?.length ? this.id + '-hint' : undefined}
+                aria-describedby={this.hint?.length || this.invalid ? this.id + '-hint' : undefined}
```
fynnfeldpausch, line 328: I think you should use the same logic that is used to show the form hint (L353)

fynnfeldpausch, line 328: This will only solve the issue for inputs, what about all the other form elements?

fynnfeldpausch (changes requested):

core/src/components/cat-input/cat-input.tsx
malikov917: done

malikov917: other elements are updated with the same logic (added aria-describedby to those which didn't have it before since hint should support a11y)

fynnfeldpausch:
This was merged within another PR. Thanks for the help.

PR #430, opened 2023-11-26
COYOFOUR-21943: enable beta releases

malikov917:
What and Why:

Our team struggle with fixing Catalyst issues.

1. We need a process for releasing fixes independent from Fynn F. 
2. We need to test Catalyst updates on Branch Deployments
3. We need to make sure Catalyst fixes are easy

AC:

1. Documentation for starting with Catalyst library is up to date, all steps are clear, no issues during set up process.
2. Additional scripts is in place which provide stable beta releases process. The process for doing beta releases is well documented.


fynnfeldpausch:
The thing with this PR is that it actually goes a step in the wrong direction. We are trying to get away from a manual release process and thus are now using release please for it. Thats the reason I am hesitant to merge it. The correct way would be to actually create some kind of CI step to create a beta release. Furthermore, what is the benefit of a beta release over a normal one?

In Addition, I think it is important to create documentation on how to link it properly so that one can test it without publication.

malikov917:
> ...Furthermore, what is the benefit of a beta release over a normal one?

We want to achieve autonomy on the development and therefore we need a process for creating builds and check them on branch deployment.

I believe having beta releases we will experience less issues with releasing (1) potentially bugged prod versions and developers would hesitate less on getting started with development of Catalyst

> The correct way would be to actually create some kind of CI step to create a beta release.
Agree. And absolutely up for it. Anyway these scripts could be used by pipelines afaik. Do we have such one for usual releases? I was thinking it's manual still for it.

> In Addition, I think it is important to create documentation on how to link it properly so that one can test it without publication.
I added the documentation, if it's not enough I can update with better description ofc

fynnfeldpausch:
> I believe having beta releases we will experience less issues with releasing (1) potentially bugged prod versions and developers would hesitate less on getting started with development of Catalyst

As said, I think the correct way to do it would be to not even possible bugged versions and instead locally link the library for local development. For this, we are missing docs.

> Agree. And absolutely up for it. Anyway these scripts could be used by pipelines afaik. Do we have such one for usual releases? I was thinking it's manual still for it.

Then we are missing the important and tricky thing, which is integrating this in the CI. Currently, I am hesitant to merge it, as manual releases whatsoever will interfere with the new release setup, which then has a lot of downsides. The auto-generted release MRs will become invalid, no GitHub releases are not created, the CHANGELOG is not properly updated and much more. So we should not encourage the creation and publication of version outside of the CI and instead find a way to build beta releases via the CI if we really need / want it.

> I added the documentation, if it's not enough I can update with better description ofc

Where? I couldn't find and docs on linking the project for development. Am I missing something?

PR #444, opened 2024-01-22
Add panelClass to CatDialogConfig type

hmezen:


angular/projects/catalyst/src/lib/dialog/dialog.service.ts
```diff
@@ -6,7 +6,7 @@ import { Observable } from 'rxjs';
 /**
  * Configuration options for modal dialogs.
  */
-export type CatDialogConfig<D = unknown> = Pick<DialogConfig<D>, 'data' | 'disableClose' | 'height' | 'role' | 'width'>;
+export type CatDialogConfig<D = unknown> = Pick<DialogConfig<D>, 'data' | 'disableClose' | 'height' | 'role' | 'width' | 'panelClass'>;
```
fynnfeldpausch, line 9: This will not be enough. It needs to be applied when opening as well.

hmezen: that is the plan

PR #462, opened 2024-02-22
feat: add attachToElement property to cat-datepicker component

glushkova91:
Problem: We need the ability to attach datepicker modal to the input element instead of the document body to control position flow.
Example: User opens datepicker in sidebar, then the user scrolls main content. As a result datepicker modal dialog is scrolled together with main content though it should be attached to sidebar and stay still. 

Solution: We can use `appendTo` config and pass catalyst datepicker component. However there is the open bug https://github.com/flatpickr/flatpickr/issues/1619 about issues with calculation of position when `appendTo` is used. Possible recommended workarounds (using `static`, css customization) doesn't work. The only possible workaround which I found was to pass custom `position` option config which accepts a function and use `floating-ui` in similar way we used it for `cat-select` for correct positioning.


core/src/components/cat-datepicker/cat-datepicker.scss
```diff
@@ -1,5 +1,7 @@
 @use 'variables' as *;
 @use 'mixins' as *;
+@import '~flatpickr/dist/flatpickr.css';
```
TimSielemann, line 3: why are those imports needed?

glushkova91: we need it because when we attach datepicker to custom dev warapper element inside shadow component (and cat-datepicker is a shadow component) then the html is encapsulated for global css. So before the datepicker modal was always inserted to the body and global css (flatpickr/dist/flatpickr.css was a part of global css) applies to calendar modal html, but it's different for html inside shadow dom and I wasn't able to find a better solution

TimSielemann: understood.

PR #512, opened 2024-05-06
feat(angular): expose some cat-scrollable props and methods in cat-dialog component

glushkova91:


angular/projects/catalyst/src/lib/dialog/dialog.component.ts
```diff
@@ -16,6 +25,26 @@ import { CatDialogHeaderComponent } from './dialog-header.component';
...
 export class CatDialogComponent implements AfterContentInit {
+  /**
+   * Flag to disable/enable overscroll behavior.
+   */
+  @Input() noOverscroll = false;
+
+  /**
+   * Flag to not fire an initial scroll event after content initialization.
+   */
+  @Input() noContentScrolledInit = false;
```
fynnfeldpausch, line 36: What is your reasoning for renaming the inputs?

fynnfeldpausch (approved):

fynnfeldpausch:
Looks good to me.

angular/projects/catalyst/src/lib/dialog/dialog.component.ts
glushkova91: Yea, probably there is no point to rename it, can't see a real scenario that we might have a scrollable header or actions. I'll rename it back

PR #564, opened 2024-09-17
fix(core): fix setting value programmatically on cat-time

geralrodrigues:
Fix an issue where the input value was not changed or been set properly when setting a value programmatically on the cat-time.
This also fix some issues on the 12 hours format where the AM/PM were not being set properly on load. When for example the time was set to PM and we were changing it programmatically to AM time, the AM/PM button was not changed.

### DIFF
--- a/core/src/components/cat-time/cat-time.tsx
+++ b/core/src/components/cat-time/cat-time.tsx
@@ -28,6 +28,8 @@ export class CatTime {
 
   @State() isAm = true;
 
+  @State() programmaticallyChanged = false;
+
   /**
    * Whether the label need a marker to shown if the input is required or optional.
    */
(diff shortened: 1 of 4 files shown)


core/src/components/cat-time/cat-time.tsx
```diff
@@ -234,22 +243,27 @@ export class CatTime {
...
+      newValue = formatIso(time);
     }
     // we need to set the input explicitly to sync the input even without a
     // rerender (if the value is not changed)
     if (this.input) {
       this.input.value = this.format(this.selectionTime, false);
     }
-    if (oldValue !== this.value) {
+    if (oldValue !== newValue) {
+      this.programmaticallyChanged = true;
```
TimSielemann, line 262: what exactly is this flag doing? The name does not really gives a hint.

geralrodrigues: It is a flag to know if the value was changed inside the component or outside the component, it is true that the name could be improved, maybe `valueChangedInternally` or `valueChangedBySync`  or `valueChangedBySelection`

geralrodrigues: changed

PR #568, opened 2024-09-23
fix(core): fix the dropdown hidden trigger initialization

geralrodrigues:
**Context**
When using a screen reader, such as VoiceOver, the state of the dropdown (reduced/expanded) is not read until interaction with the element is made. To fix this issue, it is proposed to initialize the trigger on load, even if it is still hidden.

**What to test**

- In the dropdown section, all triggers already have the a11y attributes (`aria-haspopup`, `aria-expanded`, `aria-controls`) without the need to interact with the element, even the nested dropdown (` Example user menu -> Available` and `Nested dropdown -> Click one more time)`.
- When navigating to the dropdown using a screenreader, the state (‘reduced’/‘expanded’) of the dropdown is read.

### DIFF
--- a/core/src/components/cat-dropdown/cat-dropdown.tsx
+++ b/core/src/components/cat-dropdown/cat-dropdown.tsx
@@ -18,6 +18,18 @@ let nextUniqueId = 0;
 })
 export class CatDropdown {
   private static readonly OFFSET = 4;
+  private readonly focusableSelectors = [
+    'input',
+    'select',
+    'textarea',
+    'button',
+    'a[href]',
+    '[tabindex]',
+    'iframe',
+    '[contenteditable]',
+    'audio[controls]',
+    'video[controls]'
+  ];
   private readonly id = nextUniqueId++;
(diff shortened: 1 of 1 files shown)


glushkova91:
should fix #629

core/src/components/cat-dropdown/cat-dropdown.tsx
```diff
@@ -271,6 +257,13 @@ export class CatDropdown {
     if (!trigger) {
       trigger = firstTabbable(this.triggerSlot);
     }
+    if (!trigger) {
+      elems = this.triggerSlot?.assignedElements?.() || [];
+      while (!trigger && elems.length) {
+        const elem = elems.shift();
+        trigger = findFirstTabbableIncludeHidden(elem as HTMLElement);
```
fynnfeldpausch, line 264: I think we should add a comment here why we need the special tabbable implementation here. Just to know what the issue was later.

glushkova91: agree, done

Issue #569, opened 2024-09-24
Floating elements not positioned well in dialog


geralrodrigues:
When a floating element (cat-tooltip, cat-dropdown, cat-select and cat-datepicker even it is deprecated) is opened within a dialog, the element is not positioned properly.

It only happens in the latest version of chrome, so it's very possible that it's a chrome regression bug.

**Env**: Chrome Version 129.0.6668.58 (Build oficial) (arm64) 

Additional information:
**Similar issues:** 

- https://github.com/floating-ui/floating-ui/issues/1729
- https://issues.chromium.org/issues/40843490
- https://issues.chromium.org/issues/328588659

Releases notes: https://developer.chrome.com/release-notes/129

https://github.com/user-attachments/assets/c8ac7860-25d5-420b-bce1-2e07621850ce

PR #570, opened 2024-09-26
fix(angular): fix position of popup cat-elements inside cat-dialog

geralrodrigues:
A workaround for https://github.com/haiilo/catalyst/issues/569

**What is happening?**
The floating elements of catalyst have the position property as `fixed`. When the elements are inside the dialog the position is calculated with the new container element, but it seems that the new chrome version (129) still sets the viewport as the container. That is to say, the position calculations are ok but when rendering it, the container element is the viewport instead of the dialog.

**What has been done?**
Dialog has been chosen as the new container element. 
https://developer.mozilla.org/en-US/docs/Web/CSS/position#fixed_positioning

**What to test?**
- With the Angular project running, check that the floating elements (`cat-dropdown`, `cat-date`, `cat-time`, `cat-tooltip`, `cat-select`, `cat-date`, `cat-datepicker` although it is deprecated) are positioned well within the dialog. 
- Test it with the latest version of chrome and in other browsers.

### DIFF
--- a/angular/projects/catalyst-demo/src/app/dialog/dialog.component.html
+++ b/angular/projects/catalyst-demo/src/app/dialog/dialog.component.html
@@ -11,12 +11,20 @@
     [attachToElement]="true"
   >
   </cat-datepicker>
-  <p>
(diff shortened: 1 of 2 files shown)


TimSielemann:
approved

PR #571, opened 2024-09-26
fix(core): datepicker month locale are not correct

atabaDev:
To reproduce the issue, change your local timzone to Montreal and open calendar. You'll notice that the month is not correct

### DIFF
--- a/core/src/components/cat-datepicker/cat-datepicker.locale.ts
+++ b/core/src/components/cat-datepicker/cat-datepicker.locale.ts
@@ -1,8 +1,8 @@
 import flatpickr from 'flatpickr';
-import { catI18nRegistry as i18n } from '../cat-i18n/cat-i18n-registry';
+import {catI18nRegistry as i18n} from '../cat-i18n/cat-i18n-registry';
 
 export function getHour12(language: string): boolean {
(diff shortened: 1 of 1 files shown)


core/src/components/cat-datepicker/cat-datepicker.locale.ts
```diff
@@ -23,9 +23,11 @@ function daysForLocale(language: string, weekday: 'long' | 'short' | 'narrow' =
 }
 
 function monthsForLocale(language: string, month: 'long' | 'short' = 'long') {
-  const date = new Date(0);
-  const format = new Intl.DateTimeFormat(language, { month }).format;
-  return [...Array(12).keys()].map(month => format(new Date(date.getTime()).setUTCMonth(month)));
+  return Array.from({ length: 12 }, (_, i) => {
+    const date = new Date();
+    date.setMonth(i);
+    return date.toLocaleString(language, { month: month });
```
andrei-colta-haiilo, line 29: A bit nitpicky, but you can write
`return date.toLocaleString(language, { month });`

atabaDev: Done thank you

PR #575, opened 2024-09-30
feat(core): add required marker for checkboxes

TimSielemann:


core/src/components/cat-checkbox/cat-checkbox.tsx
```diff
@@ -193,8 +199,20 @@ export class CatCheckbox {
               <polyline points="1.5 5 10.5 5"></polyline>
             </svg>
           </span>
-          <span class="label" part="label">
+          <span class={{ label: true, 'label-wrapper': !this.hasSlottedLabel }} part="label">
```
glushkova91, line 202: why don't we want to have label-wrapper class when hasSlottedLabel? at least other components don't have this condition

TimSielemann: because i want to have this backward compatible, and setting it to display flex to block might change existing usages visually...

glushkova91 (approved):

PR #591, opened 2024-10-22
fix(core): avoid endless loop on input error handling and fix wrong b…

fynnfeldpausch:
Calling `onErrorsChanged` in the willRender lifecycle method causes an endless loop. Also, setting errors="true" has no effect. Both is fixed with this PR


core/src/components/cat-tag/cat-tag.tsx
```diff
@@ -160,25 +167,19 @@ export class CatTag {
   }
 
   @Watch('errors')
-  onErrorsChanged(value?: boolean | string[] | ErrorMap) {
+  onErrorsChanged(newValue?: boolean | string[] | ErrorMap, _oldValue?: unknown, update: boolean = true) {
```
batcer, line 170: But it's called everywhere in only one matter:

```this.onErrorsChanged(this.errors, undefined, false);```

Why then other params needed, when will they be used?

fynnfeldpausch: Its a little bit hacky. The method is called via the @Watch annotation with the new and old value (and the update value missing, thus defaulting to true). So this kind of keeps the old logic, but it can also be called explicitly once a component loads (but without the update).

PR #605, opened 2024-11-15
feat(core): emit catChange on blur for cat-tag

glushkova91:


core/src/components/cat-tag/cat-tag.tsx
```diff
@@ -119,6 +119,11 @@ export class CatTag {
    */
   @Prop() tagCreationChars: string[] = [' '];
 
+  /**
+   * Whether the catChange event will be emitted when the input is blurred.
```
fynnfeldpausch, line 123: Bit of misleading docs here. It's not about the catChange event. It's more general the adding on blur

fynnfeldpausch (approved):

core/src/components/cat-tag/cat-tag.tsx
glushkova91: updated

PR #621, opened 2024-12-17
feat(core): fixes broken style for cat-time buttons

glushkova91:
8905f2b5-e11f-4567-8e89-a42e9d424bc7


glushkova91:
> How did this break? 🤔

Maybe somehow it's never worked? Not sure

PR #630, opened 2025-01-17
fix(core): cat-date-inline month display in headline

fstbraz:
This pull request fixe  the cat date inline month headline because when we switch from locale, as an example switching from DE to US, the index in the array of `locale.months.long` also changed impacting only the US customers, that will see the title showing the previous month:

Screenshot 2025-01-17 at 15 14 19
Screenshot 2025-01-17 at 10 39 12

### DIFF
--- a/core/src/components/cat-date-inline/cat-date-inline.tsx
+++ b/core/src/components/cat-date-inline/cat-date-inline.tsx
@@ -430,7 +430,7 @@ export class CatDateInline {
   }
 
   private getHeadline() {
-    return `${this.locale.months.long[this.viewDate.getMonth()]} ${this.viewDate.getFullYear()}`;
+    return `${this.viewDate.toLocaleString('default', { month: 'long' })} ${this.viewDate.getFullYear()}`;
   }
 
   private getWeekNumber(date: Date, iso8601 = true) {
--- a/core/src/components/cat-date-inline/cat-date-inline.tsx
+++ b/core/src/components/cat-date-inline/cat-date-inline.tsx
@@ -430,7 +430,7 @@ export class CatDateInline {
   }
 
   private getHeadline() {
-    return `${this.viewDate.toLocaleString('default', { month: 'long' })} ${this.viewDate.getFullYear()}`;
+    return `${this.viewDate.toLocaleString(this.language, { month: 'long' })} ${this.viewDate.getFullYear()}`;
   }
 
   private getWeekNumber(date: Date, iso8601 = true) {
--- a/core/src/components/cat-date-inline/cat-date-inline.tsx
+++ b/core/src/components/cat-date-inline/cat-date-inline.tsx
@@ -430,7 +430,7 @@ export class CatDateInline {
   }
 
   private getHeadline() {
-    return `${this.viewDate.toLocaleString(this.language, { month: 'long' })} ${this.viewDate.getFullYear()}`;
+    return `${this.locale.months.long[this.viewDate.getMonth()]} ${this.viewDate.getFullYear()}`;
   }
 
   private getWeekNumber(date: Date, iso8601 = true) {
--- a/core/src/components/cat-date-inline/cat-date-locale.ts
+++ b/core/src/components/cat-date-inline/cat-date-locale.ts
@@ -13,7 +13,8 @@ function getMonths(language: string, month: 'long' | 'short' = 'long') {
 
   return Array.from({ length: 12 }, (_, index) => {
     const fixedDate = new Date(date.getTime());
-    fixedDate.setMonth(index);
+    fixedDate.setUTCMonth(index);
+    fixedDate.setUTCHours(12);
 
     return format(fixedDate);
   });
--- a/core/src/components/cat-date-inline/cat-date-locale.ts
+++ b/core/src/components/cat-date-inline/cat-date-locale.ts
@@ -8,16 +8,8 @@ function getDays(language: string, weekday: 'long' | 'short' | 'narrow' = 'long'
 }
 
 function getMonths(language: string, month: 'long' | 'short' = 'long') {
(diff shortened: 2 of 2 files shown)


core/src/components/cat-date-inline/cat-date-inline.tsx
```diff
@@ -430,7 +430,7 @@ export class CatDateInline {
   }
 
   private getHeadline() {
-    return `${this.locale.months.long[this.viewDate.getMonth()]} ${this.viewDate.getFullYear()}`;
+    return `${this.viewDate.toLocaleString('default', { month: 'long' })} ${this.viewDate.getFullYear()}`;
```
glushkova91, line 433: should we pass this.language instead 'default'?

fstbraz: That is getting the default locale that the system is using. So probably we don't want to pass any language there

fynnfeldpausch: It totally breaks with the rest of the implementation though, which is using the generated locale in the component. If this.locale.months contains wrong values, I suggest fixing this instead of leaving the bug in the system and working around it.

fynnfeldpausch (changes requested):

core/src/components/cat-date-inline/cat-date-inline.tsx
fstbraz: @fynnfeldpausch After talking with @glushkova91, I got what was breaking. Now, we are passing the language also

fynnfeldpausch: But `this.locale.months` is still wrong. Why are we not fixing the underlying cause?

glushkova91:
> How can I reproduce the error in that function? Which locales did you pass in exactly? Because DE and US give correct results for me

@fynnfeldpausch what helped me to reproduce: 1) change system time to January 1 around 00:00 and set US timezone, e.g. Boston city. Then November instead December gonna be displayed

fynnfeldpausch:
The problem is that the local function sets the UTC date and then transforms the local date. Thats what should be fixed...

core/src/components/cat-date-inline/cat-date-locale.ts
```diff
@@ -10,7 +10,14 @@ function getDays(language: string, weekday: 'long' | 'short' | 'narrow' = 'long'
 function getMonths(language: string, month: 'long' | 'short' = 'long') {
   const date = new Date(0);
   const format = new Intl.DateTimeFormat(language, { month }).format;
-  return [...Array(12).keys()].map(month => format(new Date(date.getTime()).setUTCMonth(month)));
+
+  return Array.from({ length: 12 }, (_, index) => {
+    const fixedDate = new Date(date.getTime());
+    fixedDate.setUTCMonth(index);
```
fynnfeldpausch, line 16: I think the problem is solved by just using setMonth instead of setUTCMonth here...

fynnfeldpausch (changes requested):

core/src/components/cat-date-inline/cat-date-locale.ts
fstbraz: @fynnfeldpausch yep, solved it, thanks

glushkova91: hey guys, I already tried using setMonth instead of setUTCMonth, and seems like it breaks months headlines even more. Try to change system date/time (1.1.2025 around midnight) and Boston timezone, then months headlines are total mess

fstbraz: @glushkova91 you're right at November and February are missing for me. Going back to use the setUTCHours

fstbraz: @fynnfeldpausch is ok to go like this? Any concerns?

fynnfeldpausch: I really dislike the approach to just merge it "because it seems working" instead of really understanding and fixing the root cause. Yes it is all theoretical, but Kiribati has 14 hrs time delay and thus will also result in wrong labels. If you want to merge it, fine. I will look for a decent solution afterwards.

fynnfeldpausch: What about this? I am trying to find a solution that uses correct local dates instead of fuzzing around with UTC dates:

```
function getMonths(language: string, month: 'long' | 'short' = 'long') {
  const format = new Intl.DateTimeFormat(language, { month }).format;
  return [...Array(12).keys()].map(month => format(new Date(2000, month, 1)));
}
```

`new Date()` should return local dates afaik

fstbraz: Both solutions are acceptable and return the same output for the month calendar, at least for my manual tests. I will switch to use your proposal

Screenshot 2025-01-23 at 02 13 32

fynnfeldpausch: Seems like we need to revisit the entire implementation at some point. starting the week on tuesday seems wrong to me as well.

fynnfeldpausch (approved):

core/src/components/cat-date-inline/cat-date-locale.ts
fstbraz: Yep, I just noticed that also. Do you wanna merge this, or we block and start seeing how to improve all of this?

fynnfeldpausch: I merged it, but it would be good to create a follow up ticket

fstbraz: Ok, I will do some more testing and create one. Thanks

PR #631, opened 2025-01-20
feat(core): add updateSize method to CatTextarea component

glushkova91:


core/src/components/cat-textarea/cat-textarea.tsx
```diff
@@ -195,6 +195,14 @@ export class CatTextarea {
     this.catChange.emit(this.value);
   }
 
+  /**
+   * Update textarea size
+   */
+  @Method()
+  async updateSize(): Promise<void> {
+    autosize.update(this.textarea);
```
fynnfeldpausch, line 203: Why do we need this?

fynnfeldpausch: Is it not possible to react on the change of the value and trigger resize within the component?

PR #664, opened 2025-02-10
fix(core): do not show focus in dropdown if it was opened by mouse click

glushkova91:
**Context**
When user uses dropdown, the first element in dropdown gets focus regardless of origin. It might be opened by mouse click, by keyboard or by api and focus will be received all the time. We only had `hasInitialFocus` property to regulate this, but mainly users are interested to see focus when it's opened by keyboard and do not receive focus when it's opened by mouse.

**What has been done**

Track if click event was received from keyboard + `open` method accepts a parameter `isFocusVisible` to override internal state and pass this to `initialFocus` api function of focus-track.

### DIFF
--- a/core/src/components/cat-dropdown/cat-dropdown.tsx
+++ b/core/src/components/cat-dropdown/cat-dropdown.tsx
@@ -26,6 +26,12 @@ export class CatDropdown {
   private content!: HTMLElement;
   private trap?: focusTrap.FocusTrap;
   private isOpen: boolean | null = false;
+  /**
+   * Tracking the origin of opening the dropdown and specify if initial focus should be set.
+   * Currently we set it only when the origin is keyboard.
+   * We might not need to track this in future when focus-visible support is improved across browsers
+   */
+  private hasInitialFocus = false;
 
   /**
    * The placement of the dropdown.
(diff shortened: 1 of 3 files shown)


core/src/components/cat-dropdown/cat-dropdown.tsx
```diff
@@ -56,6 +62,7 @@ export class CatDropdown {
   /**
    * No element in dropdown will receive focus when dropdown is open.
    * By default, the first element in tab order will receive a focus.
+   * @deprecated
```
fynnfeldpausch, line 65: Why was this deprecated?

glushkova91: because using noInitialFocus property would be a bad practice from a11y perspective. We always want visible focus to jump inside of the dropdown when user uses keyboard and noInitialFocus allows to turn it off which might introduce a bug. The changes I made should cover the main reason the property was introduced for.

fynnfeldpausch: Maybe thats exactly what you shouldve written behind the @deprecated as a deprecation message ;). But not important.

fynnfeldpausch (approved):

core/src/components/cat-dropdown/cat-dropdown.tsx
glushkova91: agree and done :)

PR #670, opened 2025-02-22
feat(core): add test-id as a shorthand for data-test native attributes

fynnfeldpausch:


core/src/components/cat-checkbox/cat-checkbox.tsx
```diff
@@ -110,6 +110,13 @@ export class CatCheckbox {
    */
   @Prop() nativeAttributes?: { [key: string]: string };
 
+  /**
+   * A unique identifier for the button that is used for testing purposes. The
```
glushkova91, line 114: "button" is copied to all other component's descriptions

fynnfeldpausch: hahaha. DuDum

PR #671, opened 2025-02-22
feat(angular): add a structural directive to instantiate a basic tooltip

fynnfeldpausch:


glushkova91 (commented):
Nice! do we also want to add a support for "disabled" and "placement" inputs? We are used quite often

fynnfeldpausch:
> Nice! do we also want to add a support for "disabled" and "placement" inputs? We are used quite often

I was thinking the same but decided to keep it minimal for now. Let's see how it works and then maybe add more things. There is always the possibility to do it the old way, still.

Issue #716, opened 2025-04-11
Handle deprecated cleave.js dependecy


ddoe8:
It seems that cleave.js dependecy (used on cat-input) as been marked as deprecated (https://github.com/nosir/cleave.js). The documentation of the library mentions to use cleave-zen (https://github.com/nosir/cleave-zen) instead.

At the moment this creates an issue of using the latest features from Catalyst on search-ui repo, since updating the Catalyst to one of the latest versions breaks the search-ui build.

glushkova91:
done in #717

PR #717, opened 2025-04-11
feat(core): replace cleave.js with cleave-zen

glushkova91:


glushkova91:
fix for #716 716

core/package.json
```diff
@@ -50,7 +50,7 @@
     "@stencil/core": "4.18.3",
     "autosize": "6.0.1",
     "autosize-input": "1.0.2",
-    "cleave.js": "^1.6.0",
+    "cleave-zen": "^0.0.17",
```
fynnfeldpausch, line 53: pin it

core/src/components/cat-input/cat-input.tsx
```diff
@@ -271,16 +280,6 @@ export class CatInput {
     this.catChange.emit(this.value);
   }
 
-  /**
-   * Adds a Cleave.js mask to the input.
-   *
-   * @param options The Cleave.js options.
-   */
-  @Method()
-  async mask(options: CleaveOptions): Promise<void> {
```
fynnfeldpausch, line 280: Keep in mind that this is a breaking change. we need to build the version accordingly

core/src/components/cat-input/cat-input.tsx
```diff
@@ -211,6 +210,16 @@ export class CatInput {
    */
   @Prop() testId?: string;
 
+  /**
+   * Activates cleave-zen time mask on input
+   */
+  @Prop() timeMaskOptions?: FormatTimeOptions;
```
fynnfeldpausch, line 216: Do we want to just expose the options to the outside? I know we did it in the past, but it ties us to the library and forces us to do breaking changes in the past. Also, if we update the library, we will create a breaking change without noticing. 🤔

core/src/components/cat-input/cat-input.tsx
```diff
@@ -211,6 +210,16 @@ export class CatInput {
...
+
+  /**
+   * Activates cleave-zen date mask on input
+   */
+  @Prop() dateMaskOptions?: FormatDateOptions;
```
fynnfeldpausch, line 221: same here

fynnfeldpausch (changes requested):

core/src/components/cat-input/cat-input.tsx
glushkova91: I'd be happy to not expose it outside, but we need to specify different configuration for time and date components, and we need to update input value instantly on input event, so we should apply this formatted value in cat-input component, not in cat-time or cat-date. So configs need to be public inputs. Besides we already actively use libraries like floating-ui and expose some properties outside (though of course less 3d party dependencies we have - better for design system).

glushkova91: done

PR #737, opened 2025-05-14
feat(core): adaptive cat-tabs + stencil version update

glushkova91:
**Adapting tabs with "More" button.**

By activating _adaptive_ property, visible tabs become responsive and they are adjusted based on the parent container width. The rest of menu items, which don't fit to the parent container, are collapsed into dropdown which can be opened clicking on More button.

_activeTabAlwaysVisible_ prioritizes active button and keeps it always visible considering the tabs order.
_sticky_ prioritize a certain tab and keeps it always visible considering the tabs order.
Sticky tab has a priority over the active tab if there is not enough space for both of them.


core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -99,44 +122,174 @@ export class CatTabs {
...
-              testId={tab.testId}
-              nativeAttributes={{ ...tab.nativeAttributes }}
-              nativeContentAttributes={{ 'data-text': tab.label }}
-              data-dropdown-no-close
-            >
-              {tab.label}
-            </cat-button>
-          );
-        })}
-        <slot name="more"></slot>
```
fynnfeldpausch, line 132: removing the named slot is a breaking change. Its not a problem but we need to remember it when building the new version and updating. I guess your team is the only one using it right now

core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -99,44 +122,174 @@ export class CatTabs {
...
+                </ul>
+              </nav>
+            </cat-dropdown>
+          ) : null}
+        </div>
       </Host>
     );
   }
 
+  private adjustAdaptiveTabs() {
```
fynnfeldpausch, line 213: I am guessing that the calculation does not work when the tab style is justified? Did you try that?

fynnfeldpausch (changes requested):

core/src/components/cat-tabs/cat-tabs.tsx
```diff
@@ -99,44 +122,174 @@ export class CatTabs {
...
+                class={{
+                  'cat-tab': true,
+                  'cat-tab-active': tab.id === this.activeTab,
+                  'cat-tab-error': tab.error,
+                  'cat-tab-sticky': tab.sticky
+                }}
+                active={tab.id === this.activeTab}
+                color={tab.error ? 'danger' : tab.id === this.activeTab ? 'primary' : 'secondary'}
+                variant="text"
+                icon={tab.icon ? (tab.error ? '$cat:input-error' : tab.icon) : undefined}
```
fynnfeldpausch, line 140: edge case, but if a tab's error property is changed (or the label as well), the width will change and one would have to recalculate as well. Quick fix'd be to make an update() method available so that it can be called manually. I think you wanted to do that anyways, right?

core/src/components/cat-tabs/cat-tabs.tsx
glushkova91: calculation works pretty good with justified as well (were I broke justified when introduced tablist wrapper, will fix)

core/src/components/cat-dropdown/cat-dropdown.tsx
```diff
@@ -77,6 +77,14 @@ export class CatDropdown {
    */
   @Prop() noInitialFocus = false;
 
+  /**
+   * Whether the dropdown is open.
+   * @readonly
+   */
+  @Prop() get isOpen() {
+    return this._isOpen;
```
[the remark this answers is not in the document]
glushkova91: thanks, fixed

core/src/components/cat-tabs/cat-tabs.tsx
glushkova91: done

core/src/components/cat-tabs/cat-tabs.scss
```diff
@@ -61,11 +54,29 @@
     color: cat-token('color.theme.primary.textActive');
   }
 
+  &.cat-tab-active::after {
```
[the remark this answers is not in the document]
glushkova91: https://github.com/haiilo/catalyst/issues/738

fynnfeldpausch (approved):

PR #787, opened 2025-10-07
fix(core): in cat-form-group check horizontal attributes for children elements instead of properties

glushkova91:


core/src/components.d.ts
```diff
@@ -366,14 +366,18 @@ export namespace Components {
           * The validation errors for this input. Will render a hint under the input with the translated error message(s) `error.${key}`. If an object is passed, the keys will be used as error keys and the values translation parameters. If the value is `true`, the input will be marked as invalid without any hints under the input.
          */
         "errors"?: boolean | string[] | ErrorMap;
+        /**
+          * If the horizontal value is not provided, this fallback value is used. Can be set by form-group.
+         */
+        "fallbackHorizontal"?: boolean;
```
TimSielemann, line 372: phew i don't like the naming. But i can't imagine a better one tbh. Can we somehow remove it from public api?

TimSielemann (approved):

PR #793, opened 2025-10-14
feat(core): style cat-input of type color

TimSielemann:


core/src/components/cat-input/cat-input.tsx
```diff
@@ -375,7 +376,15 @@ export class CatInput {
               {this.icon && !this.iconRight && (
                 <cat-icon icon={this.icon} class="icon-prefix" size="l" onClick={() => this.doFocus()}></cat-icon>
               )}
-              <div class="input-inner-wrapper">
+              <div
+                class={{
+                  'has-clearable': this.clearable && !this.disabled && !this.readonly && !!this.value,
+                  'has-toggle-password': this.togglePassword && !this.disabled && !this.readonly && !!this.value,
```
glushkova91: #minor we can move both conditions to reusable private methods as a small improvement, e.g. isClearButtonShown and isTogglePasswordButtonShown

core/src/components/cat-input/cat-input.tsx
```diff
@@ -375,7 +376,15 @@ export class CatInput {
...
+                  'input-inner-wrapper': true,
+                  'type-color': this.type === 'color'
+                }}
+              >
+                {this.type === 'color' && <span class="color-value">{this.value ?? '#000000'}</span>}
```
glushkova91: theoretically this.value can be invalid and input color will be black because it's a fall back for the invalid color, but I couldn't find quickly how we can validate it and at least show some info for the user

PR #975, opened 2026-04-19
chore: configure trusted publishing

glushkova91:
The goal of this PR is to configure the release workflow for trusted publishing https://docs.npmjs.com/trusted-publishers.
Trusted publishing also doesn't support multiple release workflow files and workflow actions (https://github.com/npm/documentation/issues/1755). So I had to
1. Combine release and release-beta to one workflow file
2. Get rid of build-and-publish.yml reused workflow.


gluecksmelodie (approved):
LGTM 🔐

PR #987, opened 2026-05-22
feat(core): implement scoped cat-icons via attachSet + set name

glushkova91:


core/src/components/cat-icon/cat-icon.tsx
```diff
@@ -34,10 +39,21 @@ export class CatIcon {
    */
   @Prop({ attribute: 'a11y-label' }) a11yLabel?: string;
 
+  connectedCallback() {
+    this.setName = this.findSetName();
+  }
```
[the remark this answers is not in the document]
glushkova91: neither should ever happen

core/src/components/cat-icon/cat-icon.tsx
```diff
@@ -48,4 +64,8 @@ export class CatIcon {
       ></span>
     );
   }
+
+  private findSetName(): string | undefined {
+    return findClosest(`[${CAT_ICON_SET_ATTR}]`, this.el)?.getAttribute(CAT_ICON_SET_ATTR) ?? undefined;
```
fynnfeldpausch: Isn't `?? undefined` what happens anyways with the null coalescing? But it does not hurt either I guess...

fynnfeldpausch (dismissed):

core/src/components/cat-icon/cat-icon.tsx
glushkova91: we still need null coalescing here bcs getAttribute can return null or undefined and better to return consistent not defined value, btw replaced with null

PR #996, opened 2026-06-05
feat: disable autocomplete by default

fynnfeldpausch:


core/src/components/cat-input/cat-input.tsx
```diff
@@ -414,6 +414,7 @@ export class CatInput {
                     'has-toggle-password': this.togglePassword && !this.disabled && !this.readonly && !!this.value
                   }}
                   autocomplete={this.autoComplete}
+                  data-1p-ignore={this.autoComplete === 'off' ? '' : undefined}
```
ddoe8, line 417: [minor]: In theory we could this.autoComplete.toLowerCase() === 'off', just to safeguard, someone setting autoComplete "OFF" or "Off". The risk is low i guess

ddoe8:
@fynnfeldpausch guess cat-date and cat-time don't have this issue and therefor the data-1p-ignore is not needed on those

ddoe8 (approved):
Approved with minor comments

core/src/components/cat-input/cat-input.tsx
fynnfeldpausch: true, but I think its a problem on the dev side

PR #1007, opened 2026-06-25
COYOFOUR-26737 cat-select show tooltip for labels

ianovalena:


core/src/components/cat-select/cat-select.tsx
```diff
@@ -703,26 +748,33 @@ export class CatSelect {
...
-                  onInput={this.onInput.bind(this)}
-                  value={!this.multiple ? this.state.term : undefined}
-                  placeholder={this.placeholder}
-                  disabled={this.disabled || this.state.isResolving}
-                ></input>
+                <cat-tooltip
+                  trigger="hover"
+                  content={!this.multiple && this.state.selection.length ? this.state.selection[0].render.label : ''}
+                  placement="top"
+                  disabled={this.multiple || !this.state.selection.length || !this.singleValueTruncated}
```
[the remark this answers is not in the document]
ianovalena: Cat tooltip uses display contents, so the existing styles are not affected

core/src/components/cat-select/cat-select.tsx
glushkova91, line 1: would be nice to add an example with tooltip to the demo (index.html file)

core/src/components/cat-select/cat-select.tsx
```diff
@@ -998,6 +1069,7 @@ export class CatSelect {
         activeSelectionIndex: -1
       });
     }
+    this.needsInputTruncationCheck = true;
```
glushkova91, line 1072: do we need to check trancation on deselect?

core/src/components/cat-select/cat-select.tsx
```diff
@@ -659,39 +697,46 @@ export class CatSelect {
...
-                        class={{
-                          pill: true,
-                          'select-no-open': true,
-                          'select-option-active': this.state.activeSelectionIndex === i
-                        }}
-                        role="option"
-                        aria-selected="true"
-                        id={`select-${this.id}-selection-${i}`}
+                      <cat-tooltip
+                        key={item.item.id}
```
glushkova91: do we need it?

core/src/components/cat-select/cat-select.tsx
```diff
@@ -838,6 +890,7 @@ export class CatSelect {
           class="select-option"
           id={`select-${this.id}-option-${i}`}
           aria-selected={isOptionSelected ? 'true' : 'false'}
+          key={item.item.id}
```
glushkova91, line 893: is it used?

core/src/components/cat-select/cat-select.tsx
```diff
@@ -1089,6 +1161,29 @@ export class CatSelect {
...
+    const keys = new Set<string>();
+    this.dropdown?.querySelectorAll<HTMLElement>('.select-option-label[data-option-key]').forEach(el => {
+      if (el.scrollHeight > el.clientHeight) {
+        keys.add(el.dataset.optionKey!);
+      }
+    });
+    this.truncatedOptionKeys = keys;
+  }
+
+  private checkInputTruncation(): void {
```
glushkova91, line 1174: maybe it makes sense to rename to checkTriggerTruncation since we also truncate selected options in the trigger/label (if I understood correctly)

glushkova91 (dismissed):

core/src/components/cat-select/cat-select.tsx
ianovalena: Yes, because we save indexes of pills that should have a tooltip

core/src/components/cat-select/cat-select.tsx
ianovalena: I would keep input, because it's inside or on top of input

core/src/components/cat-select/cat-select.tsx
ianovalena: Yes, it's fixing a bug with a wrong checkmark in cases when the retrieve function has no delay

core/src/components/cat-select/cat-select.tsx
ianovalena: I think no

core/src/components/cat-select/cat-select.tsx
glushkova91: maybe we can just set truncatedPillIndices to empty set and singleValueTruncated to false then

core/src/components/cat-select/cat-select.tsx
glushkova91: would be nice to have a bit more details about that since key is not an official html attribute

core/src/components/cat-select/cat-select.tsx
ianovalena: But it deselects a single item, not the entire list

core/src/components/cat-select/cat-select.tsx
ianovalena: Added a comment about it, but overall, it's a stencil DOM directive

glushkova91 (approved):

